PR #5, opened 2020-05-13
Add support for wiki pricing, optimize parsing, add testing framework

rm-you:
* Wiki parsing for item auto-links and historical pricing
* Update parsing logic to use regex for reliability
* Remove node_modules (probably shouldn't be checked in)
* Set up testing with the Jest framework


tests.js
```diff
@@ -21,7 +20,7 @@ function parseLog(text, logServer) {
     // console.log(words);
     //test if is auction
     if (words[1] === 'AUCTIONS,') {
-        // client.streamAuction(text.replace('||', '|'), "GREEN");
+        fakeStreamAuction(text.replace('||', '|'), logServer); // for testing purposes, don't use client.streamAuction
```
rm-you, line 23: I'm actually curious why you needed to de-duplicate pipes, are they common in auction messages?

jamesjamail: Not very common but apparently || is discord markdown for spoiler or something like that.  There are other discord markdown I should probably be checking for too, but spoiler was really annoying for the "stream".

rm-you: Ahh, that makes sense! So just in case someone did this, like `/auc WTS FBSS 16k || Something Else 1k` it wouldn't do dumb stuff

rm-you:
I think it actually already *is* in .gitignore, but you force-added it at some point. The only real change is to delete the whole directory but I can do that separately to not make this commit huge / confusing.

rm-you:
So, your first checkout/pull with this code may freak out a bit because of any changes you have locally to node_packages. Hopefully you know how to deal with that (I would delete the directory completely, do a hard-reset with git, and then do the pull/checkout). From then on it should be fine.

logParser.js
```diff
@@ -29,57 +32,38 @@ setInterval(() => {
...
-        //trim single quotes
-        words[2] = words[2].slice(1);
-        words = filterWTS(words);
-        auction = words.join(' ');
-        if (words.length > 2) {
+        const auction_user = auction_text[1];
+        const auction_contents = auction_text[2];
+        const auctionWTS = filterWTS(auction_contents);
+        if (auctionWTS) {
             itemList.forEach(({item_name, user_id, user_name, price, server}) => {
```
rm-you, line 53: Humorously, this is exactly the same problem I solved with Aho-Corasick for the wiki lookup parsing. This works fine with a small itemList (few users creating watches) and it's possible it will never be a scale problem for us, but I will look into switching it to do a linear-time search using the same algorithm. No reason to do `O(n)` matches when it could be `O(1)`.

rm-you: Ok, I looked through this logic again and A-C might not work, or at least not with the current way the watches are stored. Since there could be multiple watches with the same keyword, or parts of keywords, and A-C stops matching for a section once it finds a match, it would interfere with finding all possible matches for all users. There may not be a great way around this using A-C.

Fortunately, I think the scale should be relatively small. We may also want to implement a watch quota so users can only watch some (reasonable) amount of keywords, something like 30 to start. That'd ensure that someone couldn't macro watching literally one million things and DoS the bot.

rm-you:
Ah, I still need to figure out how to tweak the discord client to not show embeds on messages. Should be possible I hope...

rm-you:
Hmm, maybe the solution is actually to make each of these actually *be* embeds using 
https://discord.js.org/#/docs/main/stable/class/MessageEmbed
rather than just a text link? Could make them look a lot nicer, and even pull the image link from the wiki for the item if it exists...

logParser.js
jamesjamail: Great idea!

jamesjamail:
Really great work.  Sorry it took me so long to review these.

jamesjamail:
> So, your first checkout/pull with this code may freak out a bit because of any changes you have locally to node_packages. Hopefully you know how to deal with that (I would delete the directory completely, do a hard-reset with git, and then do the pull/checkout). From then on it should be fine.

I'm curious if it would be possible to delete the node modules in this repo, so that when the guy running the bot only has to do git pull/npm install.  Would that work?  Just trying to make the instructions for the guy as simple as possible since he isn't very familiar with git or node.

PR #8, opened 2020-05-31
Testing Refactor

jamesjamail:
-see comments in the spaghetti

TODO:

-verify watch feature is working correctly
-confirm tailing two log files works
-if easy, I'd love to have the wiki scraper on watch responses.
-SHIP IT

### DIFF
--- a/client.js
+++ b/client.js
@@ -1,8 +1,6 @@
 const Discord = require('discord.js');
 const logger = require('winston');
-//settings.test for test server 
-const auth = require('./settings.test.json');
-// const auth = require('./settings.json');
+const settings = require('./settings.json');
 const helpMsg = '\n\n***TunnelQuestBot Help***\n***NOTE:***\n-All commands begin with an exclamation mark (\"!\").\n-Arguments listed in carats (\"<\" \">\") should be replaced by your input data.\n-Item names are not case sensitive.\n-You may enter prices in pp or kpp (ex: 1100pp or 1.1k).\n-Parser will not detect aliases (ex: watching "Thick Banded Belt" will not detect "TBB"), however this is a future goal.\n\n***COMMANDS***\n!help   (displays available commands)\n!add watch: <item>, <at or below this price>, <server>   (starts a watch based on enetered parameters - watches expire after 7 days.  Price is optional)\n!end watch: <item>   (ends a currently running watch)\n!end all watches   (ends all currently running watches)\n!extend all watches   (extends your current watches another 7 days)\n!show watch: <item>   (lists details for a watch for entered item - if no item is provided, lists details for all watches)\n!show watches   (lists details for all watches)\n\n ***TIPS***\n-You use !add watch to update an existing watch if you wish to modify the price and/or reset the 7 day expiration timer'
 const db = require('./db.js');
 const {fetchAndFormatAuctionData} = require("./wikiHandler");
@@ -16,7 +14,10 @@ logger.level = 'debug';
 
 // Initialize Discord Client
 var bot = new Discord.Client();
-
+const TOKEN = settings.discord.token;
+const COMMAND_CHANNEL = settings.discord.command_channel;
+const BLUE_STREAM_CHANNEL = settings.discord.BLUE.stream_channel;
+const GREEN_STREAM_CHANNEL = settings.discord.GREEN.stream_channel;
 
 
 bot.on('ready', () => {
@@ -25,9 +26,9 @@ bot.on('ready', () => {
 
 bot.on('message', function (message) {
 
-    // ignore bots and self and chat rooms
+    // ignore bots and non-command rooms
     // console.log(message);
-    if (message.author.bot || message.channel.id === '675891646235279386' || message.channel.id === '673793154729771028' || message.channel.id === '672512233435168784'){
+    if (message.author.bot || message.channel.id !== COMMAND_CHANNEL){
         return;
     }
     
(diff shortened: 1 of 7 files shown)


client.js
```diff
@@ -1,6 +1,8 @@
 const Discord = require('discord.js');
 const logger = require('winston');
-const auth = require('./settings.json');
+//settings.test for test server 
+const auth = require('./settings.test.json');
```
rm-you, line 4: Really no reason to need a different settings file in code, right? Just swap out the actual file, the idea of a settings file is you don't have to make code changes. But, I guess if that's your workflow, that's fine, since I assume this'll be removed before this merges anyway.

client.js
```diff
@@ -162,14 +164,24 @@ function streamAuction (auction_user, auction_contents, server) {
     let channelID;
 
     if (server === "GREEN") {
-        channelID = "672512233435168784";
+        // channelID = "672512233435168784";
+        //for test server user...
+        channelID = "716521067388338207";
```
rm-you, line 169: Eugh, forgot to move channelID over to the discord settings file. Need to do that.

client.js
```diff
@@ -162,14 +164,24 @@ function streamAuction (auction_user, auction_contents, server) {
...
     }
-    // console.log(msg, server, channelID);
+    // console.log("bot.channels.get", bot.channels.fetch("716521067388338207"));
 
     auction_contents = auction_contents.replace(/[|]+/g, '|');
     fetchAndFormatAuctionData(auction_user, auction_contents, server).then(formattedAuctionMessage => {
-        // in the real code this would do:
-        bot.channels.cache.get(channelID).send(formattedAuctionMessage)
+
+        bot.channels.fetch("716521067388338207")
```
rm-you, line 176: Need to actually USE `channelID` :D

logParser.js
```diff
@@ -13,10 +13,11 @@ if (require.main === module) {
     const db = require('./db.js');
     const settings = require('./settings.json');
     const tail = require('tail');
-    const itemList = new Set();
+    // const itemList = new Set();
+    let itemList = [];
```
rm-you, line 17: Does it actually not work with a set? That'd be sad. :(

logParser.js
```diff
@@ -25,7 +26,13 @@ if (require.main === module) {
...
         db.upkeep();
         db.getWatches((results) => {
-            itemList.add(results);
+            //I think it might be best to just have itemList = results here for a couple of reasons:
+            //a) I like the idea of a set to eleminate searching duplicate data, but I'm pretty sure objects with the same properties do not deeply equate since they have separate memory references (confirmed - just got a 100 pings for 1 watch)
+            //b) it's impossible to have duplicates anyway, as each result object has a username
+            //c) watches expire after 7 days and users can update them at any time - if we just continue adding to itemList, users may get alerts for out of date info.
+
+            //results.forEach((result) => itemList.add(result)) //I refactored your code a bit here to see if it would work (results is an array so need to iterate over it to add to a set), but issue a above just makes a continuing growing list
+            itemList = results;
```
rm-you, line 35: Yeah ok, that's fair, I just kinda sprinted through this section since I couldn't run/test it without a test DB, so I didn't think to hard about what that list was. So, the problem then is that this changes the value of the pointer for itemList instead of just mutating the actual data -- which means the pointer that was passed to parseLog is invalid. To make this work, itemList needs to be defined ourside of the scope of this `if` block, I THINK? Unless JS doesn't treat `if` blocks as their own scope, which is definitely possible....

logParser.js
```diff
@@ -45,30 +52,37 @@ function parseLog(text, itemList, logServer, client) {
     //test if is auction
     const auction_text = text.match(AUC_REGEX);
     if (auction_text) {
-        client.streamAuction(text.replace(/[|]+/g, '|'), logServer);
         const auction_user = auction_text[1];
         const auction_contents = auction_text[2];
+        client.streamAuction(auction_user, auction_contents.replace(/[|]+/g, '|'), logServer);
```
rm-you, line 57: I think we should remove the `replace(/[|]+/g, '|')` here and do sanitization inside the stream function now, as there will be more to do.

logParser.js
```diff
@@ -45,30 +52,37 @@ function parseLog(text, itemList, logServer, client) {
...
+                    let logPrice = parsePrice(filteredAuction, item_name.length);
+                    if (price === -1 && logPrice === null) {
+                        let msg = {userId: user_id, userName: user_name, itemName: item_name, sellingPrice: logPrice, seller: auction_user, server: server, fullAuction: text};
+                        console.log("match found - no price requirement", logPrice, price, msg)
+                        outgoing.push(msg);
+                    }
+                    else if (logPrice && logPrice <= price || price === -1) {
+                        let msg = {userId: user_id, userName: user_name, itemName: item_name, sellingPrice: logPrice, seller: auction_user, server: server, fullAuction: text};
+                        console.log("Meets price criteria", logPrice, price, msg)
+                        outgoing.push(msg);
```
rm-you, line 82: Did you change anything in this section or is it all just de-indented one level? Trying to spot differences and it's hurting my eyes, lol

logParser.js
```diff
@@ -45,30 +52,37 @@ function parseLog(text, itemList, logServer, client) {
...
                     }
                 }
-            ) 
+            })
```
rm-you, line 85: Wait what

logParser.js
jamesjamail: Sorry, I just uncommented some previous console.log statements.  Only minor change is to the conditionals for checking price.

logParser.js
jamesjamail: Wow this was confusing to look at a diff of.  The only change is that "})" is now on one line instead of two (because they are both opened on the same line in the forEach call above.

logParser.js
jamesjamail: This is the way I had it previously and never got any bugs.  Isn't itemList defined much further above already?

client.js
jamesjamail: channelID.toString()

I still don't know why.  Maybe a quirk of Discord.js.

client.js
jamesjamail: It's ignored so shouldn't be a problem.

logParser.js
rm-you: No, I'd moved it into this if-block -- it's only defined a few lines up, inside this scope. But, if this works, then maybe it's fine (and if-blocks in JS don't have their own scope).

jamesjamail:
I probably should have merged this before making a new PR.  Feel free to handle it however you'd like- my latest PR is running very smoothly.

rm-you:
Closing because the other PR will contain all changes.

PR #9, opened 2020-06-02
tested parsing both servers and fixed errors in wiki handler

jamesjamail:
Working well now.  Will let it run overnight to track any errors.

-added tail for blue server
-modified the way stream channel id was being imported
-removed unnecessary vars
-added some clarifications to log paths in settings.example
-reset interval to fetch data every minute instead of every 10 seconds
-changed forEach function in wiki handler to for loop due to errors

Mtil brought up that he wanted to migrate the bot to another computer sometime this week, so figure that will be a good opportunity to start with a fresh git clone using this version.

### DIFF
--- a/client.js
+++ b/client.js
@@ -1,8 +1,6 @@
 const Discord = require('discord.js');
 const logger = require('winston');
-//settings.test for test server 
-const auth = require('./settings.test.json');
-// const auth = require('./settings.json');
+const settings = require('./settings.json');
 const helpMsg = '\n\n***TunnelQuestBot Help***\n***NOTE:***\n-All commands begin with an exclamation mark (\"!\").\n-Arguments listed in carats (\"<\" \">\") should be replaced by your input data.\n-Item names are not case sensitive.\n-You may enter prices in pp or kpp (ex: 1100pp or 1.1k).\n-Parser will not detect aliases (ex: watching "Thick Banded Belt" will not detect "TBB"), however this is a future goal.\n\n***COMMANDS***\n!help   (displays available commands)\n!add watch: <item>, <at or below this price>, <server>   (starts a watch based on enetered parameters - watches expire after 7 days.  Price is optional)\n!end watch: <item>   (ends a currently running watch)\n!end all watches   (ends all currently running watches)\n!extend all watches   (extends your current watches another 7 days)\n!show watch: <item>   (lists details for a watch for entered item - if no item is provided, lists details for all watches)\n!show watches   (lists details for all watches)\n\n ***TIPS***\n-You use !add watch to update an existing watch if you wish to modify the price and/or reset the 7 day expiration timer'
 const db = require('./db.js');
 const {fetchAndFormatAuctionData} = require("./wikiHandler");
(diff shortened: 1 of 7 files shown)


settings.example.json
```diff
@@ -1,4 +1,23 @@
 {
-  "token":  "DISCORD_API_TOKEN",
-  "logFilePath": "C:\\Program Files (x86)\\Sony\\EverQuest\\Logs\\eqlog_Auclog_P1999Green.txt"
+  "logFilePath": {
+    "BLUE": "C:\\Program Files (x86)\\Sony\\EverQuest\\Logs\\eqlog_CHARNAME_project1999.txt",
```
jamesjamail, line 3: Blue has a slightly different ending to log file

wikiHandler.js
```diff
@@ -33,13 +35,14 @@ async function fetchAndFormatAuctionData(auction_user, auction_contents, server)
 
 function formatPriceMessage(item, data) {
     const price_points = [];
-    for (let interval in data[1]) {
-        price_points.push(`[*${interval}d*] ${data[1][interval]}`);
-    }
-    const price = data[0] !== undefined ? `**${data[0]}pp** ` : '';
-    return `${item} ${price}${price_points.join(" / ")}`;
```
jamesjamail, line 40: data[1] is an object so a just replaced your forEach with a for loop.

client.js
```diff
@@ -14,18 +14,18 @@ logger.level = 'debug';
...
     logger.info(`Logged in as ${bot.user.tag}!`);
 });
 
 bot.on('message', function (message) {
 
-    // ignore bots and self and chat rooms
+    // ignore bots and non-command rooms
     // console.log(message);
-    if (message.author.bot || message.channel.id === '675891646235279386' || message.channel.id === '673793154729771028' || message.channel.id === '672512233435168784'){
+    if (message.author.bot || message.channel.id !== COMMAND_CHANNEL){
```
jamesjamail, line 28: This actually does not allow direct messages, think we will have to do it the other way unless discord.js has a meessage.channel.type prop we can check

wikiHandler.js
rm-you: heh that just undoes the change I was trying to do, sadly (get away from that type of for-loop). Maybe I can fix it differently...

db.js
```diff
@@ -1,5 +1,5 @@
 const { Client } = require('pg');
-const settings = require('./settings.json');
+const settings = require('./settings.test.json');
```
rm-you, line 2: Really trying to not change this in code -- for example, right now, if we merged this PR, it would force EVERYONE to change the name of their settings file. Just keep multiple files and overwrite `settings.json` with the one you want to use

settings.example.json
rm-you: ah, k

wikiHandler.js
rm-you: I guess it's just fine for now.

client.js
rm-you: Fixed using type.

Issue #10, opened 2020-06-03
Item parsing can incorrectly flag partial matches


rm-you:
Item names that are sub-strings of longer item names will match incorrectly. See Amber/Jasper:
```
[**Berbank**] is auctioning:
**WTS**: `WTS Golden Star Amulet, Golden Amber Earring, Jasper Gold Earring,Golden Cat Eye Bracelet 20p each (I have multiple copies), Spell: Gravity Flux 40p, Spell: Insipid Weakness 10p`
<http://wiki.project1999.com/Golden_Star_Amulet> [*30d*] 33 ± 37 / [*90d*] 28 ± 21
<http://wiki.project1999.com/Amber> [*30d*] 1 ± 0 / [*90d*] 1 ± 0
<http://wiki.project1999.com/Golden_Amber_Earring> [*30d*] 25 ± 3 / [*90d*] 26 ± 6
<http://wiki.project1999.com/Jasper> [*30d*] 5 ± 0 / [*90d*] 4 ± 1
<http://wiki.project1999.com/Jasper_Gold_Earring> [*30d*] 25 ± 6 / [*90d*] 26 ± 6
<http://wiki.project1999.com/Golden_Cat_Eye_Bracelet> **20pp** [*30d*] 24 ± 2 / [*90d*] 25 ± 6
<http://wiki.project1999.com/Gravity_Flux> **40pp** [*30d*] ? ± ? / [*90d*] 170 ± 0
<http://wiki.project1999.com/Insipid_Weakness> **10pp** [*30d*] 23 ± 10 / [*90d*] 25 ± 10
```

The actual item is Golden Amber Earring / Jasper Gold Earring.
Amber and Jasper are also items on the wiki, so they matched as well.

Possible solution:
After compiling matches, locate each item name in the auction string and make note of the start/end characters to form a range. Then, remove items with ranges that fall completely within the range of another item.

PR #27, opened 2020-06-14
No longer match on partial item names

rm-you:
If a match is completely included in a separate match, discard it.

Resolves: #10


rm-you:
Wait and merge this after #26 merges, and this has been rebased.

Issue #18, opened 2020-06-12
Bot should inform people commands in general channel aren't recognized


jamesjamail:
Many users try to enter commands in the general channel, then assume the bot is not working when they get no response.

Others seeing commands in general chat encourages them to do the same, and the cycle continues.

When a user issues a message starting with ! in the general channel, Bot should message the channel (not the user directly, so that others can see it too):

"Sorry, commands are not recognized in this channel.  Please use public_command_space or message me directly.  Thank you."

PR #33, opened 2020-06-15
Bot deletes messages outside command channel and alerts user

jamesjamail:
Completes "Bot should inform people commands in general channel aren't recognized"

Also made minor tweaks to !help message

Resolves: #18

### DIFF
--- a/client.js
+++ b/client.js
@@ -1,35 +1,7 @@
 const Discord = require('discord.js');
 const logger = require('winston');
 const settings = require('./settings.json');
-const helpMsg = '\n\n' +
-    '**TunnelQuestBot Help**\n' +
-    '***NOTE:***\n' +
-    ' • All commands begin with an exclamation mark (\"!\").\n' +
-    ' • Arguments listed in `code blocks` should be replaced with your input data.\n' +
-    ' • Commas *are* actually required.\n' +
-    ' • Item names are not case sensitive.\n' +
-    ' • You may enter prices in pp or k (ex: 1100pp or 1.1k).\n' +
-    ' • Parser will not detect aliases (ex: watching "Thick Banded Belt" will not detect "TBB"), however this is a future goal.\n' +
-    '\n' +
-    '***COMMANDS:***\n' +
-    '!help' +
-    '   →   Displays available commands.\n' +
-    '!add watch: `item`, `maximum price`, `server`' +
-    '   →   Starts a watch based on entered parameters - watches expire after 7 days.  Price is optional.\n' +
-    '!end watch: `item`, `server`' +
-    '   →   Ends a currently running watch.\n' +
-    '!end all watches' +
-    '   →   Ends all currently running watches.\n' +
-    '!extend all watches' +
-    '   →   Extends your current watches another 7 days.\n' +
-    '!show watch: `item`, `server`' +
-    '   →   Lists details for a watch for entered item - if no arguments are provided, behaves as *!show watches*.\n' +
-    '!show watches' +
-    '   →   Lists details for all watches.\n' +
-    '\n' +
-    '***TIPS:***\n' +
-    ' • You can use `!add watch` to update an existing watch if you wish to modify the price and/or reset the 7 day expiration timer.\n' +
-    ' • To report a problem or request a feature, talk to us in #feedback_and_ideas, or create an issue here: https://github.com/jamesjamail/TunnelQuestBot/issues'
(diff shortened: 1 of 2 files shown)


client.js
```diff
@@ -13,19 +13,19 @@ const helpMsg = '\n\n' +
     '\n' +
     '***COMMANDS:***\n' +
     '!help' +
-    '> Displays available commands.\n' +
+    '   →   Displays available commands.\n' +
```
rm-you, line 16: Ah, ">" is specifically for quoting, it is special discord formatting. I BELIEVE it should look better than a unicode arrow? But I'll look at both and see what I think shortly.

client.js
```diff
@@ -51,8 +51,9 @@ bot.on('ready', () => {
 
 bot.on('message', function (message) {
     // console.log(message);
-    // ignore bots or messages from any text channel that isn't the Command Channel
-    if (message.author.bot || (message.channel.type === 'text' && message.channel.id !== COMMAND_CHANNEL)){
```
rm-you, line 55: Now that we don't short-circuit on messages in other channels, won't it hit the `else` below on messages without "!" and then spam channels when people talk?

client.js
```diff
@@ -159,6 +160,11 @@ bot.on('message', function (message) {
                 message.author.send('Sorry, I didn\'t recognized that command.  Please check your syntax and try again. ' + helpMsg);
                 break;
         }
+        // message the general channel that commands are not recognized in this channel if command detected:
+        if (!message.author.bot && message.channel.type === 'text' && message.channel.id !== COMMAND_CHANNEL){
+            bot.channels.cache.get(COMMAND_CHANNEL).send(`Hi <@${message.author.id}>, I received your command but deleted it from the \`general\` channel to keep things tidy.  In the future, please use this channel instead or send me a direct message.  Thanks!`)
+            message.delete();
+        }
     }
     else {
```
rm-you, line 169: I would expect this else block to activate on normal chat in #general now?

client.js
```diff
@@ -159,6 +160,11 @@ bot.on('message', function (message) {
                 message.author.send('Sorry, I didn\'t recognized that command.  Please check your syntax and try again. ' + helpMsg);
                 break;
         }
+        // message the general channel that commands are not recognized in this channel if command detected:
+        if (!message.author.bot && message.channel.type === 'text' && message.channel.id !== COMMAND_CHANNEL){
+            bot.channels.cache.get(COMMAND_CHANNEL).send(`Hi <@${message.author.id}>, I received your command but deleted it from the \`general\` channel to keep things tidy.  In the future, please use this channel instead or send me a direct message.  Thanks!`)
```
rm-you, line 165: Do we want to spam the command channel with this? Or should we just PM the user directly? I also don't necessarily like the idea of *publically* shaming users for making a mistake. Deleting the message has the advantage of hiding the mistake quickly, but then we announce about it loudly... :/

client.js
jamesjamail: Oh, I didn't notice.  Let me know.

client.js
jamesjamail: You're right, fixed.

client.js
jamesjamail: This is the same as your comment above I believe? If so, fixed.

client.js
jamesjamail: I liked how it works when you are @'ed from the channel you are suppose to use.  If you think there's a shame factor though we can make it a PM.

rm-you: Yeah, I suppose it's fine. It's more like an FYI, I suppose. There's technically other cool stuff we could do, like "click the checkmark reaction to acknowledge this message" and then delete that too... but lots of extra work for no real reason :D

client.js
rm-you: Yeah, IMO it looks better with the quoting, if you want to change it back 👍

jamesjamail:
> I'm good with this once you swap back to the line-quote `>`

It's ready captain.

client.js
jamesjamail: Done!

client.js
```diff
@@ -1,7 +1,7 @@
 const Discord = require('discord.js');
 const logger = require('winston');
 const settings = require('./settings.json');
-const { helpMsg, welcomeMsg } = require('./messages');
+const { helpMsg, welcomeMsg } = require('./messages')
```
rm-you, line 4: RIP semicolon

Issue #17, opened 2020-06-12
Welcome new users


jamesjamail:
Discord.js has an event for when new users join the server.  It would be a helpful feature to do two things when this happens:

1) Bot sends users a direct message:

"Hello ${user}!  I am TunnelQuestBot, your helpful gnome assistant.  Please allow me to make buying and selling items easier so you can finally start tipping for ports.

I watch auctions on both Blue and Green servers.  If you're in the market for a new sword, you can enter the following command as a direct message to me, or in the public_command_space channel:

!add watch: rusty bastard sword, green

If you only have 5pp to spend, you can enter a price criteria:

!add watch: rusty bastard sword, 5pp, green

Whenever I see a match, I'll send you a direct message with all the pertinent info.

Watches last 7 days before they expire.

For more information, try the !help command."

2) Bot messages General Chat channel, "Welcome to the server, ${user}" as a simple greeting.

I think as we progress it would be fun to make the bot persona be a snarky gnome as it's fun for users and encourages them to actually read the message instead of skim it.

PR #34, opened 2020-06-15
Server Greeting

jamesjamail:
-added server greeting to general channel
-bot sends welcome message directly to new users
-moved messages to another file

Resolves: #17

### DIFF
--- a/client.js
+++ b/client.js
@@ -1,37 +1,9 @@
 const Discord = require('discord.js');
 const logger = require('winston');
 const settings = require('./settings.json');
-const helpMsg = '\n\n' +
-    '**TunnelQuestBot Help**\n' +
-    '***NOTE:***\n' +
-    ' • All commands begin with an exclamation mark (\"!\").\n' +
-    ' • Arguments listed in `code blocks` should be replaced with your input data.\n' +
-    ' • Commas *are* actually required.\n' +
-    ' • Item names are not case sensitive.\n' +
-    ' • You may enter prices in pp or k (ex: 1100pp or 1.1k).\n' +
-    ' • Parser will not detect aliases (ex: watching "Thick Banded Belt" will not detect "TBB"), however this is a future goal.\n' +
-    '\n' +
-    '***COMMANDS:***\n' +
-    '!help' +
-    '> Displays available commands.\n' +
-    '!add watch: `item`, `maximum price`, `server`' +
-    '> Starts a watch based on entered parameters - watches expire after 7 days.  Price is optional.\n' +
-    '!end watch: `item`, `server`' +
-    '> Ends a currently running watch.\n' +
-    '!end all watches' +
-    '> Ends all currently running watches.\n' +
-    '!extend all watches' +
-    '> Extends your current watches another 7 days.\n' +
-    '!show watch: `item`, `server`' +
-    '> Lists details for a watch for entered item - if no arguments are provided, behaves as *!show watches*.\n' +
-    '!show watches' +
-    '> Lists details for all watches.\n' +
-    '\n' +
-    '***TIPS:***\n' +
-    ' • You can use `!add watch` to update an existing watch if you wish to modify the price and/or reset the 7 day expiration timer.\n' +
-    ' • To report a problem or request a feature, talk to us in #feedback_and_ideas, or create an issue here: https://github.com/jamesjamail/TunnelQuestBot/issues'
+const { helpMsg, welcomeMsg } = require('./messages');
 const db = require('./db.js');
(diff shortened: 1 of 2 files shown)


client.js
```diff
@@ -49,6 +21,18 @@ bot.on('ready', () => {
...
 });
 
+//invite in case I can't find server after leaving https://discord.gg/Q8nzh5
+
+//server greeting for users who join
+bot.on('guildMemberAdd', (member) => {
+    let guild = member.guild; // Reading property `guild` of guildmember object.
+    let memberTag = member.user.tag; // GuildMembers don't have a tag property, read property user of guildmember to get the user object from it
+    if (guild.systemChannel){ // Checking if it's not null
+        guild.systemChannel.send(`Welcome to the server, ${memberTag}!`);
```
rm-you, line 31: So, what exactly IS our goal here? If it's really just to say "hi!" I don't know if this is really worth it. We might want to have it PM them off the bat, maybe with the helptext and a quick welcome blurb, so they can have some info about how things work?

client.js
```diff
@@ -49,6 +21,18 @@ bot.on('ready', () => {
...
+    }
+    bot.users.cache.get(member.user.id).send(`**Hi ${memberTag}!**\n\n` + welcomeMsg);
```
rm-you, line 33: Ah nevermind, here it is like 2 lines down, lol.

messages.js
```diff
@@ -0,0 +1,44 @@
...
+    '!show watches' +
+    '> Lists details for all watches.\n' +
+    '\n' +
+    '***TIPS:***\n' +
+    ' • You can use `!add watch` to update an existing watch if you wish to modify the price and/or reset the 7 day expiration timer.\n' +
+    ' • To report a problem or request a feature, talk to us in #feedback_and_ideas, or create an issue here: https://github.com/jamesjamail/TunnelQuestBot/issues'
+
+
+const welcomeMsg = 'I am TunnelQuestBot, your helpful gnome assistant. Please allow me to make buying and selling items easier so you can finally start tipping for ports.\n\n' +
+    'I watch EC auctions on both Blue and Green servers. If you\'re in the market for a new sword, you can enter the following command as a direct message to me, or in the \`public_command_space channel\`:\n\n' +
```
rm-you, line 33: Maybe put "channel" outside of the code block? and make it #public_command_space (I assume we can't link to it via channel-id in a PM? did you try? maybe it randomly works lol)

messages.js
```diff
@@ -0,0 +1,44 @@
...
+    '\`!add watch: rusty bastard sword, green\`\n\n' +
+    'If you only have 5pp to spend, you can enter a price criteria:\n\n' +
+    '\`!add watch: rusty bastard sword, 5pp, green\`\n\n' + 
+    'Whenever I find a match, I\'ll send you a direct message with all the pertinent info.\n\n' +
+    'Watches last 7 days before they expire.\n\n\n' +
+    'You can also check out our Tunnel Stream channels.\n\n' +
+    'Auction message are displayed at the top of each post with links to the wiki items beneath.  Hovering over the item name displays historical pricing data.\n\n\n' +
```
rm-you, line 40: Could be more specific and say "historical pricing data courtesy of the P1999 Wiki"? but this is fine

messages.js
jamesjamail: What's the functionality you are aiming for?

rm-you: I was just curious if it's possible to make a clickable channel link to a server, in a DM. Not sure it's possible.

rm-you: regardless, at least changing:
as a direct message to me, or in the \`public_command_space channel\`:\n\n'
to:
as a direct message to me, or in the \`#public_command_space\` channel:\n\n'
would be good I think?

jamesjamail: You can make a clickable link to a channel from a message.  I'm down to refactor to that eventually, though I'd rather integrate this PR the way it is now because I don't have much bandwidth today.

I'll add the '#' though!

messages.js
```diff
@@ -0,0 +1,44 @@
...
+    ' • Parser will not detect aliases (ex: watching "Thick Banded Belt" will not detect "TBB"), however this is a future goal.\n' +
+    '\n' +
+    '***COMMANDS:***\n' +
+    '!help\n' +
+    '> Displays available commands.\n' +
+    '!add watch: `item`, `maximum price`, `server`\n' +
+    '> Starts a watch based on entered parameters - watches expire after 7 days.  Price is optional.\n' +
+    '!end watch: `item`, `server`\n' +
+    '> Ends a currently running watch.\n' +
+    '!end all watches' +
```
rm-you, line 18: missed a \n here

PR #39, opened 2020-06-23
Added SQL schema to repo

jamesjamail:
-schema.txt added to root level of project directory


rm-you:
I believe this is superseded by #42 since that will add v001 schema and v000 will no longer be necessary to have in tree, as the migrations will exist for 000->001 and nothing->001?

rm-you:
Yeah, superseded by #42 which has now merged.

PR #41, opened 2020-08-27
Implement Redis for caching fetch requests

jamesjamail:
-pretty straight forward.  Set cached data to expire after 24 hours.
-[link to windows installer for Redis](https://github.com/MicrosoftArchive/redis/releases/download/win-3.0.504/Redis-x64-3.0.504.msi)

### DIFF
--- a/wikiHandler.js
+++ b/wikiHandler.js
@@ -114,8 +114,7 @@ async function getWikiPricing(item_url, server) {
     cache.get(item_url, (err, cached_data) => {
         if (err) console.error(err);
         //use cached data if available
-        if (cached_data) {
-            console.log('cached worked for ', item_url)
+        if (cached_data !== null) {
             parsePage(cached_data, server)
         } else {
             //otherwise fetch new data
@@ -132,21 +131,6 @@ async function getWikiPricing(item_url, server) {
                 .catch(console.error)
         }
     })
-
-    return fetch(item_url)
-        .then(response => response.text())
-        .then(text => {
-            const $ = cheerio.load(text);
-            const auc_data = $(`#auc_${server} .eoTable3 td`).contents();
-            let price_data = {};
-            if (auc_data !== undefined && auc_data[0] !== undefined) {
-                price_data[30] = auc_data[0].data.trim();
-            }
-            if (auc_data !== undefined && auc_data[1] !== undefined) {
-                price_data[90] = auc_data[1].data.trim();
-            }
-            return price_data;
-        }).catch(console.error)
 }
 
 async function findWikiData(auction_contents, server) {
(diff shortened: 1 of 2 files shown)


wikiHandler.js
```diff
@@ -86,10 +93,46 @@ function formatPrice(price) {
     }
 }
 
+function parsePage(page, server) {
```
rm-you, line 96: This is duplicated code now -- now that it's a function, can we clean that up below?

wikiHandler.js
```diff
@@ -86,10 +93,46 @@ function formatPrice(price) {
...
+                        parsePage(text, server)
+                    })
+                })
+                .catch(console.error)
+        }
+    })
+
     return fetch(item_url)
         .then(response => response.text())
         .then(text => {
```
rm-you, line 138: Seems like the function parsePage above was pulled from here? So could this be replaced with a call to that somehow? We don't want duplicate codepaths for the same thing if possible.

wikiHandler.js
```diff
@@ -86,10 +93,46 @@ function formatPrice(price) {
...
     // Translate server name to "capitalized" form, eg: GREEN -> Green
     server = server[0] + server.slice(1).toLowerCase();
 
+    //check cache before fetching
+    cache.get(item_url, (err, cached_data) => {
+        if (err) console.error(err);
+        //use cached data if available
+        if (cached_data) {
+            console.log('cached worked for ', item_url)
+            parsePage(cached_data, server)
```
rm-you, line 119: So we're storing the raw page response in cache? I guess this is fine, but it'd be much smaller to cache the actual result we want, right? And then we'd only have to run parsePage once before storing, not every time we fetch.

wikiHandler.js
```diff
@@ -86,10 +93,46 @@ function formatPrice(price) {
...
+                        if (err) console.error(err);
+                        //parse for price
+                        parsePage(text, server)
+                    })
+                })
+                .catch(console.error)
+        }
+    })
+
     return fetch(item_url)
```
rm-you, line 136: Wait, so we fetch here ... but this is also copied somewhere else -- and again, should be a single code-path, this seems duplicated also.

wikiHandler.js
rm-you, line 119: Second comment here -- after we do the parsePage, does it ... not return the parsed data? Do we fall through to like 136 below and run fetch() a second time?

jamesjamail: 100%.  Was thinking that this morning.  Now it's much more proper.

jamesjamail: Oof. Yeah, my bad- remnants of your code i was pulling apart.  Deleted now.

PR #48, opened 2023-03-23
Added WTB Functionality

cwwp671:
Hey, I have attempted to make WTB functionality for you guys. Let me know what you think.

    /src/client/commands/watch.js - Added an option to select WTB or WTS for the watch command.
    /src/client/executors.js - Modified the watch() function to handle the WTB option.
    /src/parser/logParser.js - Updated the parseLog() function to handle WTB messages.
    /src/client/clientHelpers.js - Updated the watchBuilder() function to display the watch type (WTS or WTB) in the generated embeds.
    /src/client/client.js - Modified the pingUser() function to handle both WTS and WTB cases and customize the message sent to the user based on the watch type.
    /src/db/db.js - Updated the addWatch() function to handle the WTB case.

### DIFF
--- a/src/client/client.js
+++ b/src/client/client.js
@@ -55,6 +55,7 @@ bot.on("ready", () => {
 // commands during development and publish them to global commands when they're
 // ready for public use.
 bot.commands = new Collection();
+console.log("bot.commands type:", typeof bot.commands);
 const commands = [];
 for (const file of commandFiles) {
   const command = require(`./commands/${file}`);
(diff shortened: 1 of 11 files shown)


rm-you:
Sorry, just noticed that you actually made this PR. 😭 
Thanks for your contribution!

This generally looks sane, but I would still like to carefully review what the string changes / final output looks like with our test deployment.

If you wouldn't mind rebasing this on top of the `node18_refactor` branch, the tests will all now run properly, and we can make sure everything works as expected. I can do that as well after we merge it, but likely would speed things up if you have time to do that and maybe add a little more testing now that all the tests run again. Also, ping @Devs in discord if you do this, so we know you're actively working on it and can pay more attention. 👍

Issue #49, opened 2023-03-29
zumaik's animation not showing up


smasherprog:
<img width="610" alt="image" src="https://user-images.githubusercontent.com/3393733/228646280-6691abb0-3e7a-45e8-b069-d6b6fc80e859.png">

if you search you will see its in the body of the message, but not as a found item.

smasherprog:
Is this repo maintained?

rm-you:
Yes, it is, but we don't necessarily check the Issues page often -- report bugs in discord in the feedback channel!
As for this bug -- the author of the other issue appears to understand the problem, and this is a common one -- many items have discrepancies in their naming when they use apostrophes, between what the wiki believes and what is true in-game.
I'll look at fixing it soon.

rm-you:
@smasherprog Should be fixed! Sorry for the delay. :)

rm-you:
Following up, the change was merged but the new version didn't roll out properly, so we're stuck on some of these issues until the refactor is done, hopefully within a week or two.

PR #52, opened 2023-05-21
Fix items with apostrophes/backticks

rm-you:
Duplicate items with apostrophes/backticks so that any version of their name will register (it's very unpredictable which ones are correct in the wiki and which are not).

Resolves: #49


jamesjamail (approved):
This looks good- thanks!

PR #58, opened 2023-11-20
Replace the parser implementation (based on the classic one)

rm-you:


src/lib/parser/parser.ts
```diff
@@ -32,70 +32,42 @@ export class AuctionParser {
...
-			}
-		}
-
-		// Check if the word includes any of the WTB auction types
-		for (const auctionType of wtbAuctionTypes) {
-			if (word.includes(auctionType)) {
-				return AuctionTypes.WTB;
+	// Thanks rici from StackOverflow for saving me time!
+	// Based on https://stackoverflow.com/a/30472781
+	private composeRanges(ranges: { start: number; end: number }[]) {
```
rm-you, line 37: I do kinda wish I could rewrite this function to be less "magical", it's the one part about this that is incomprehensible unless you know what it's doing already. I'm sure it could be written in a more sane style.

src/lib/parser/parser.ts
```diff
@@ -32,70 +32,42 @@ export class AuctionParser {
 		return processedMsg;
 	}
 
-	private wordIsAuctionType(word: string) {
-		const wtsAuctionTypes = ['WTS', 'SELLING'];
```
rm-you, line 36: There's admittedly no replacement for "SELLING/BUYING" (my code only checks WTS/WTB) but I am also not sure I've EVER seen that spelled out, so ...

src/lib/parser/parser.ts
```diff
@@ -104,113 +76,108 @@ export class AuctionParser {
...
-				)
-			) {
-				// If not, reset unknownItemString and return early
-				unknownItemString = '';
-				return;
+		const wtb_match = preprocessedMessage.match(/WTB/);
+		const wts_match = preprocessedMessage.match(/WTS/);
+		if (wtb_match && wts_match) {
+			// eslint-disable-next-line @typescript-eslint/ban-ts-comment -- linter is wrong here
+			// @ts-ignore -- if both ANDs pass, then both are defined
```
rm-you, line 135: This is the first time I've ever seen TypeScript just be WRONG. Very weird.

tsconfig.json
```diff
@@ -39,15 +39,15 @@
...
+    "allowJs": true,                                  /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
     // "checkJs": true,                                  /* Enable error reporting in type-checked JavaScript files. */
     // "maxNodeModuleJsDepth": 1,                        /* Specify the maximum folder depth used for checking JavaScript files from 'node_modules'. Only applicable with 'allowJs'. */
 
     /* Emit */
     // "declaration": true,                              /* Generate .d.ts files from TypeScript and JavaScript files in your project. */
     // "declarationMap": true,                           /* Create sourcemaps for d.ts files. */
     // "emitDeclarationOnly": true,                      /* Only output d.ts files and not JavaScript files. */
-    // "sourceMap": true,                                /* Create source map files for emitted JavaScript files. */
+    "sourceMap": true,                                /* Create source map files for emitted JavaScript files. */
```
rm-you, line 50: This is useful for debugging.

tsconfig.json
```diff
@@ -39,15 +39,15 @@
     // "noResolve": true,                                /* Disallow 'import's, 'require's or '<reference>'s from expanding the number of files TypeScript should add to a project. */
 
     /* JavaScript Support */
-    // "allowJs": true,                                  /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
+    "allowJs": true,                                  /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
```
rm-you, line 42: Need this for that ahocorasick lib. :(

src/lib/parser/parser.ts
```diff
@@ -104,113 +76,108 @@ export class AuctionParser {
...
-					buying.push({ item, price });
+			// Figure out price
+			const price_match = Array.from(
+				segment.segment.matchAll(
+					/(?<price>[0-9]+(\.[0-9]+)?)(?<kpp>k?p{0,2})/gi,
+				),
+			);
+			if (price_match.length > 0) {
+				const single_price_match = price_match[price_match.length - 1];
+				let price: number = +single_price_match[1];
```
rm-you, line 159: Annoyingly, the regex makes nice named groups, but TypeScript won't let me use them because they're "dynamic" and not defined clearly in the type for a RegexMatch. I'm not sure how to fix that...

src/lib/parser/parser.ts
```diff
@@ -1,20 +1,20 @@
-import { Trie } from './trieSearch';
-import inGameItemsObject from '../gameData/items.json';
+import { consolidatedItemsAndAliases } from '../gameData/consolidatedItems';
 import { ItemType } from '../streams/streamAuction';
 
+// Oldschool JS lib works the best of all the options :(
```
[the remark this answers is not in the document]
jamesjamail: Glad we had a similar experiences 🤣 I felt like I was going crazy trying to get all this popular modules to work

src/lib/parser/parser.ts
jamesjamail: it happens pretty regularly unfortunately, but I'm sure we can extend your parser to handle that too.

src/lib/parser/parser.ts
jamesjamail: What are you wanting to do here? I may be able to help

src/lib/parser/parser.ts
```diff
@@ -131,9 +131,9 @@ export class AuctionParser {
 		const wtb_match = preprocessedMessage.match(/WTB/);
 		const wts_match = preprocessedMessage.match(/WTS/);
 		if (wtb_match && wts_match) {
-			// eslint-disable-next-line @typescript-eslint/ban-ts-comment -- linter is wrong here
-			// @ts-ignore -- if both ANDs pass, then both are defined
-			if (wtb_match.index < wts_match.index) {
+			// Typescript barks wtb_match and wts_match might be undefined, even though
+			// they are tested above.  ! is used to assert non-nullness
+			if (wtb_match.index! < wts_match.index!) {
```
jamesjamail, line 136: Hate it when TypeScript is dumb.  Probably a little more graceful to use ! to assert non-nullness here.

jamesjamail:
Okay I relent, this is a much better approach.  There are some indexing bugs in the parser I wrote using the trie search, but even if I resolved those it would never match items that are joined together without a space, which makes your approach so much more robust.  I wish I had reached out to you when I was struggling to get the aho corasick modules to work before I went down the rabbit hole of a trie search by word.

2 Questions:

1) I think we can just replace your regex for WTS/WTB to include "buying" and "selling" too?  As in...

```
const wtb_match = preprocessedMessage.match(/\b(WTB|BUYING)\b/i);
const wts_match = preprocessedMessage.match(/\b(WTS|SELLING)\b/i);
```

2) Is there a way to keep track of the auction string that did *not* match any items/spells/aliases and return that separately from known items in the buying and selling arrays?  As in...

```
{
  buying: {
      knownItems: [{ item: cloak of flames, price: undefined }],
      unknownItems: [{item: port to gfay, price: 50 }]
     },
   selling: {
      knownItems: [{ item: fungus covered scale tunic, price: 40000 }],
      unknownItems: [{item: blacksmithing services for hire, price: undefined }]
   }
}
```

If a user enters a watch for "jboots mq", it won't be a match from aho corasick.  Ideally we could parse the unknown auction contents within the parser so that it gets cached along with the rest of the auction data, which would save an additional regex match on the auction content if the watch is not a "known" item.

PR #63, opened 2024-03-08
Toggle State Button Interactions

jamesjamail:


rm-you (approved):
I THINK this is good now -- but since I was the last one to make changes, you should probably take a look again and approve as well. :D

jamesjamail:
> I THINK this is good now -- but since I was the last one to make changes, you should probably take a look again and approve as well. :D

This looks great- thank you so much.  I haven't had a chance to run it locally or try webstorm but this is good to merge and have Mtil update IMHO.

PR #65, opened 2024-03-28
Fix watch cache, add /version

rm-you:


src/lib/parser/monitorLogs.ts
```diff
@@ -54,7 +54,6 @@ export function monitorLogFile(server: Server) {
 		flushAtEOF: true,
 		useWatchFile: true,
 	});
-	const watchedItemsForThisServer = state.watchedItems[server];
```
rm-you, line 57: Doing this caching outside the tail function meant this reloads the watch state exactly once at boot time. I am not sure exactly why the return object from this isn't a true pointer? Something with how node works? It seems like it is additive but will never drop old references or something? IDK node well enough to know why this was behaving the way it does, but I verified in testing that this fixes it.

PR #66, opened 2024-04-26
Don't refresh unwatched items on "global refresh" (/list button)

rm-you:


jamesjamail (approved):
Looks wonderful, thanks so much!

PR #68, opened 2024-06-22
Bug Fix - Watch Block (and related improvements)

jamesjamail:


src/lib/watchNotification/watchNotification.ts
```diff
@@ -62,12 +62,22 @@ export async function shouldUserBeNotified(
 		}
```
rm-you, line 62: Uhh what about this loop? Did you mean to remove it? I think you reproduced it below but didn't remove the bad forEach version

jamesjamail: nice callout- fixed

rm-you (approved):
Looks good now :D

PR #71, opened 2024-06-22
Should not have added discord deps to dev

rm-you:
Also fix a discord.js change to the SlashCommand type


src/types.d.ts
```diff
@@ -1,5 +1,5 @@
 import {
-	SlashCommandBuilder,
+	SlashCommandOptionsOnlyBuilder,
```
rm-you, line 2: I guess this changed in discordjs 14.15.x?

package.json
```diff
@@ -51,7 +51,6 @@
     "@types/node-fetch": "^2.6.6",
     "@types/node-forge": "^1.3.7",
     "@typescript-eslint/eslint-plugin": "^6.6.0",
-    "discord.js": "^14.15.3",
```
rm-you, line 54: This can't be here or it will prune discord.js during the image creation XD

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "tunnel-quest-bot",
   "version": "1.1.2",
-  "lockfileVersion": 2,
```
rm-you, line 4: I just regenerated the entire package-lock file for the first time in a while, and this is what it did -- I guess the new version doesn't require the dependencies section? Or was that something we added manually?

package.json
jamesjamail: weird- idk how discordjs got added as a dev dep

jamesjamail (approved):
LGTM

package.json
rm-you: Me, I added it by accident 😂

PR #72, opened 2024-06-23
One more spot needs to be SlashCommandOptionsOnlyBuilder

rm-you:


rm-you:
@jamesjamail can you pull this down and like... build it and make sure it works for you? because I swear I did last time too and it worked fine, but obviously this seems necessary (though I am unclear as to why other spots don't need to use this type)

jamesjamail:
Working well

PR #96, opened 2026-08-06
Harden runtime recovery, add test suite, and upgrade to Prisma 7

rm-you:
## Summary

This PR makes the bot testable, more resilient to transient failures, and ready for Prisma 7. The work started from production pain: a network or database hiccup could crash the process, Docker would restart it, and log monitoring would drop for every server until things stabilized. Several of those failure modes also had no automated coverage, so regressions were only visible in production.

The branch adds a two-tier Vitest suite (500+ unit tests, 50+ integration tests against real Postgres and Redis), fixes the bugs that suite surfaced, upgrades Prisma to v7, and hardens runtime recovery so the bot fails loudly when it should and retries when it can.

### Reliability and runtime behavior

- Stop transient failures from taking down the whole process: await async command/autocomplete handlers, isolate notification fan-out with `Promise.allSettled`, and make error reporting itself non-fatal.
- Throttle Discord error reporting (15-minute dedupe by stack + 10/minute ceiling) so handled failures do not flood the error channel.
- Retry Discord login and startup initialization on transient dependency failures; exit on uncaught exceptions so Docker can restart a corrupted process instead of leaving a disconnected-but-healthy container.
- Fix a production boot bug where scanning `build/handlers/*.js` invoked loader modules as callables and skipped `client.login()` entirely.
- Release notification debounce keys after transient send/build failures, but keep them for permanent closed-DM errors (`50007`).

### Correctness fixes surfaced by tests

- Parser: attribute WTS/WTB per item from preceding text; prefer suffixed prices over quantity digits.
- Message builders: respect Discord per-embed limits and the 6,000-character aggregate limit across all embeds in a message; batch large auction streams into multiple sends.
- Notifications: claim debounce keys before sending; skip watches deleted between index refreshes.
- Commands/UI: fix silent no-replies, wrong unsnooze copy, empty DM states, missing try/catch in `version`, and `extendAllWatches` resurrecting inactive watches.
- Prisma 7: adapt lazy user creation to the new P2003 constraint path (`meta.driverAdapterError.cause.constraint`).
- Buttons: use atomic `SET EX NX` for interaction dedupe.
- DB: add index on `Watch.active`.

### Test infrastructure

- Unit tier: mocked Discord/Prisma/Redis, shared factories, coverage thresholds (79/75/70/79).
- Integration tier: Testcontainers Postgres + Redis, table truncation between tests, dbExecutor and higher-order function coverage.
- Extracted `commandLoader` / `eventLoader` for testability; fixed the one test that accidentally booted the real Discord client.

### Tooling, CI, and deployment

- Upgrade to Prisma 7 with `@prisma/adapter-pg`, generated client in `src/prisma/generated`, and `prisma.config.ts` + `dotenv-expand`.
- CI on PRs: lint, coverage-threshold unit tests, TypeScript build, integration tests, and a non-pushing Docker build. Image publish remains push-to-`main` only.
- Replace bash-only Dockerfile CMD with `docker-entrypoint.sh` (LF-pinned via `.gitattributes`), add `.dockerignore`, and narrow runtime image to what `migrate deploy` actually needs.
- Remove dead code and unused deps (`axios`, `node-forge`); replace deprecated `ephemeral` with `MessageFlags.Ephemeral`.
- Batch public command-space cleanup with `bulkDelete`.

## Test plan

- [x] `npm run lint`
- [x] `npm run test:coverage` (503 unit tests, thresholds met)
- [x] `npm run test:integration` (53 integration tests)
- [x] `npm run build`
- [x] `docker build` locally
- [x] GitHub Actions: `test` and `integration` jobs passing on PR

## Deployment notes

- Requires running migrations (`20260805040000_add_watch_active_index`).
- Prisma 7 expects `DATABASE_URL` to be fully expanded before client init; `.env` interpolation is now handled via `dotenv-expand` in app entrypoints and `prisma.config.ts`.
- No intentional behavior change for normal command usage; biggest user-visible wins are fewer missed notifications during outages and fewer Discord API rejections on long watch lists / large auctions.


jamesjamail:
<3

PR #97, opened 2026-08-07
Follow-up to #96: harden Prisma FK detection and untangle two module boundaries

jamesjamail:
# Harden Prisma FK detection and untangle two module boundaries

Follow-up to #96. Three commits, plus two more addressing review.

## The bug fix

`isForeignKeyViolationError` only read Prisma 7's `driverAdapterError.cause.constraint`. Prisma has relocated this detail before, and each move silently breaks lazy user creation: a missed FK violation just rethrows, so a user's very first command fails.

`attemptAndCreateUserIfNeeded` also becomes generic. Dropping `Promise<any>` restores real return types at both call sites — `upsertWatchSafely` returns `Watch` again, `insertPlayerLinkSafely` returns `string` — and lets the file-wide `no-explicit-any` disable go. `isForeignKeyViolationError` now narrows from `unknown` rather than taking `any`.

## Refactors

**`consumeReportBudget`** (was `hasExceededReportRate`). The old name read as a predicate while mutating counters that must only be touched once per attempt. The existing throttle test in `errors.test.ts` — 12 errors, expects 10 sends — guards the inversion.

**Named handler exports.** `registerCommandHandlers` / `registerEventHandlers` replace `require()` calls in `index.ts`, so renaming either fails the build instead of at boot. Motivated by the boot-time handler failure in #96.

**`getEnvironmentVariable`** moves out of `streamAuction` into `lib/helpers/env.ts`. The parser tests were stubbing it through a whole-module `vi.mock` factory, which silently replaces every export of that module; a precise mock is far less likely to rot.

## Review follow-ups

**Moved the `getEnvironmentVariable` tests** into `src/lib/helpers/env.test.ts` alongside the helper. The variable they exercise is now a neutral name rather than `SERVERS_BLUE_STREAM_CHANNEL_CLASSIC_ID`, which implied a coupling that no longer exists. Adds the two cases the original lacked (a set value is returned; an empty value is rejected like a missing one) and restores state in a `finally`, so a failed assertion can't leak a deleted variable into later tests.

**Dropped the v5/v6 FK fallbacks and the message fallback.** `prisma`, `@prisma/client` and `@prisma/adapter-pg` are pinned together at `^7.9.1`, so the flat `meta.field_name` / `meta.constraint` shapes can't occur without a deliberate downgrade — the three fixtures asserting them were testing inputs this codebase cannot produce.

The fallbacks also weren't what protected against the next relocation. `higherOrderFunctions.itest.ts` is: it drives a real `P2003` against real Postgres through the same adapter production uses, so a moved field turns the build red before anything ships. And matching `error.message` meant trusting Prisma's prose format while the comment above it declined to trust the structured fields — the weaker of the two assumptions.

Kept the `unknown` + `isRecord` typing, both driver-adapter shapes (`index` and `fields`, since non-Postgres drivers report bare column names), and every negative test — including the `BlockedPlayerByWatch` case, which is the one that could actually bite.

## Verification

530 unit tests, 55 integration tests, lint and typecheck clean. The integration suite needs Docker running.

## Note

#103 is stacked on this branch and should merge second.

### DIFF
--- a/src/handlers/Command.ts
+++ b/src/handlers/Command.ts
@@ -2,8 +2,8 @@ import { Client } from 'discord.js';
 import { join } from 'path';
 import { loadSlashCommands, registerSlashCommands } from './commandLoader';
 
-module.exports = (client: Client) => {
+export function registerCommandHandlers(client: Client) {
 	const slashCommandsDir = join(__dirname, '../lib/commands/slashCommands');
 	const slashCommands = loadSlashCommands(slashCommandsDir, client);
 	void registerSlashCommands(slashCommands);
-};
+}
--- a/src/handlers/Event.ts
+++ b/src/handlers/Event.ts
@@ -2,7 +2,7 @@ import { Client } from 'discord.js';
 import { join } from 'path';
 import { loadEvents } from './eventLoader';
 
-module.exports = (client: Client) => {
+export function registerEventHandlers(client: Client) {
 	const eventsDir = join(__dirname, '../events');
 	loadEvents(eventsDir, client);
-};
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,6 +12,8 @@ export const client = new Client({
 	partials: [Partials.Message, Partials.Channel, Partials.Reaction], //	needed for handling interactions from DM's
 });
 import { SlashCommand } from './types';
+import { registerCommandHandlers } from './handlers/Command';
+import { registerEventHandlers } from './handlers/Event';
 import { config } from 'dotenv';
 import { expand } from 'dotenv-expand';
 import {
@@ -41,11 +43,12 @@ process.on('uncaughtException', (error) => {
 client.slashCommands = new Collection<string, SlashCommand>();
 client.cooldowns = new Collection<string, number>();
 
-// 	Listed explicitly rather than scanned from handlers/: the scan also invoked
+// 	Called explicitly rather than scanned from handlers/: the scan also invoked
 // 	the loader modules, which export named functions instead of a callable, and
-// 	the resulting throw skipped the login() call below.
-require('./handlers/Command')(client);
-require('./handlers/Event')(client);
+// 	the resulting throw skipped the login() call below. Imported rather than
+// 	require()d so a rename of either handler fails the build instead of at boot.
+registerCommandHandlers(client);
+registerEventHandlers(client);
 
 // 	Once logged in discord.js reconnects on its own, but a failure during the
 // 	initial login is fatal. On container start we frequently lose the race with
--- a/src/lib/content/messages/messageBuilder.ts
+++ b/src/lib/content/messages/messageBuilder.ts
@@ -17,11 +17,8 @@ import {
 	formatPriceNumberToReadableString,
 	isSnoozed,
 } from '../../helpers/watches';
-import {
-	AuctionData,
-	ItemType,
-	getEnvironmentVariable,
-} from '../../streams/streamAuction';
+import { AuctionData, ItemType } from '../../streams/streamAuction';
+import { getEnvironmentVariable } from '../../helpers/env';
 import { getImageUrlForItem } from '../../helpers/images';
 import { getWikiUrlFromItem } from '../../helpers/wikiLinks';
 import {
--- a/src/lib/helpers/env.test.ts
+++ b/src/lib/helpers/env.test.ts
@@ -1,0 +1,49 @@
+import { describe, it, expect } from 'vitest';
+import { getEnvironmentVariable } from './env';
+
+//	A name with no meaning to the rest of the app: the helper is a generic
+//	lookup, so pinning it to a stream channel would imply a coupling that no
+//	longer exists now that it lives outside streamAuction.
+const NAME = 'A_TEST_ONLY_VARIABLE';
+
+function withVariable(value: string | undefined, run: () => void) {
+	const original = process.env[NAME];
+	if (value === undefined) delete process.env[NAME];
+	else process.env[NAME] = value;
+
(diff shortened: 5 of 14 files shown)


rm-you (commented):
Three nicely separated commits, and the rebase onto current `main` is clean — it also dropped the `concurrently` downgrade that was in the earlier version. I ran it locally: 531 passing.

Overall I'm happy to merge this. One small ask, one suggestion I'll leave to your judgment, and one follow-up I'll file separately rather than pile on here.

## What I like

Making `attemptAndCreateUserIfNeeded` generic is the highest-value change in the PR. Dropping `Promise<any>` restores real return types at both call sites — `upsertWatchSafely` returns `Watch` again, `insertPlayerLinkSafely` returns `string` — and lets the file-wide `no-explicit-any` disable go. The `unknown` + `isRecord` guard is the same improvement applied to the error path.

`consumeReportBudget` is a good catch. The old name read as a predicate while mutating counters that must only be touched once per attempt. The inversion is exact (`!(x <= MAX)` is `x > MAX`), and the existing throttle test in `errors.test.ts` — the one that fires 12 errors and expects 10 sends — guards it, so it can't silently drift.

The named handler exports are well motivated given the boot-time handler failure in #96. I also checked the hoisting question, since `import` pulls the handler modules above the `client` creation and both loaders import `gracefullyHandleError` from `errors.ts`, which imports `client` back. It's safe: that cycle already existed via the direct `errors` import in `index.ts`, CJS late-binds the property access, and nothing in that graph reads `client` at module scope. Worth saying out loud because it's the kind of thing that would only fail at runtime.

Pulling `getEnvironmentVariable` out of `streamAuction` is right, and the test benefit is bigger than the commit message claims: the parser tests were stubbing it through a whole-module `vi.mock` factory, which silently replaces every export of that module. A precise mock is far less likely to rot.

## One ask

The `getEnvironmentVariable` tests stayed behind in `src/lib/streams/streamAuction.test.ts` (the `describe` block near the top) while the source moved to `helpers/env.ts`. Could you move it into a new `src/lib/helpers/env.test.ts`? While you're in there, the assertion uses `SERVERS_BLUE_STREAM_CHANNEL_CLASSIC_ID`, which should probably become a generic name now that the helper isn't stream-specific.

## One suggestion, your call

On the FK matcher in `higherOrderFunctions.ts`: I follow the reasoning, but I think the `meta.field_name` and `meta.constraint` candidates are unreachable. We pin `prisma`, `@prisma/client`, and `@prisma/adapter-pg` all at `^7.9.1`, so the flat-`meta` shapes can't occur without a deliberate downgrade.

More to the point, I don't think the fallbacks are what protects us from the next relocation — `higherOrderFunctions.itest.ts` is. It drives the real v7 path against real Postgres through the same adapter production uses in `src/prisma/init.ts`, so if Prisma moves the field, CI goes red before anything ships. The fallbacks only soften a failure that already got past a red build.

The `error.message` fallback also sits a little awkwardly with its own premise: it is the loosest matcher, and it stays safe only as long as Prisma's prose format is stable — which is the assumption the comment right above it declines to make about the structured fields.

I'm not worried about the risk, to be clear. All four models with a `discordUserId` FK (`Watch`, `BlockedPlayerByWatch`, `BlockedPlayer`, `PlayerLink`) want the same remedy, so broad matching is semantically correct rather than sloppy, and a false match costs one idempotent `user.upsert` plus a retry that fails identically — the first attempt threw before writing anything. Your `BlockedPlayerByWatch` negative test covers the one case that could actually bite.

So: not blocking. I'd trim the two v5/v6 branches and the three fixtures that only assert shapes we can't produce, and keep the typing and the negative-case tests, which are the parts earning their keep.

## Two description quibbles

"Reads as permission rather than a negated failure" — the call site is `if (!consumeReportBudget(now))`, so it is still negated; the old form wasn't. The real win is the name signaling mutation, which is plenty on its own.

"No longer depend on streaming internals they do not exercise" — both parser tests still mock `../streams/streamAuction` for `streamAuctionToAllStreamChannels`. It went from one mock to two, just better factored.

Separately, the `bulkDelete` comment fix in `removeMessagesFromCommandSpace.ts` is accurate (Discord's default fetch limit is 50, bulk accepts 100) but unrelated to the rename commit it is riding in. Not worth a re-roll.

## Follow-up I'll file separately

`index.ts` declares `export const client = new Client(...)` above the import block, which misleads about load order since imports hoist regardless — and it is what creates the `index -> errors -> index` cycle. Extracting the client into `src/client.ts` would make the order honest and break the cycle properly. Predates this PR, so I'll open an issue instead of expanding scope here.

rm-you (approved):
Both points addressed — thanks for taking them.

The `getEnvironmentVariable` tests moved to `src/lib/helpers/env.test.ts`, and you went further than I asked: the value-returned and empty-string-is-missing cases weren't there before, and the `finally`-restoring helper means a failed assertion can't leak a deleted variable into the rest of the run. Better than what it replaced.

Dropping the fallbacks came out well too, and writing the reasoning into the comment rather than just deleting the code is the right call — the next person to hit a Prisma upgrade needs to know the itest is the guard, not rediscover it.

Verified locally: 530 passing across 77 files, which is exactly the three removed fixtures plus three new env tests minus the one relocated. CI green.

One non-blocking nit: the header comment in `higherOrderFunctions.test.ts` still reads "Prisma has relocated the violated-constraint detail between major versions. Each shape below shipped in a real release; missing one means lazy user creation stops working" — but the two shapes left are both Prisma 7 driver-adapter variants, so the cross-version rationale no longer matches the list underneath it. Worth a sentence trim whenever you next touch the file; not worth a push on its own.

Approving. I'll review #103 separately.

PR #103, opened 2026-08-10
Dx improvements

jamesjamail:
# Developer experience: setup, tooling, and CI

Makes this repo something a stranger can clone and contribute to. No runtime behaviour changes to the bot itself.

> **Stacked on #97.** Merge that one first — this branch is rebased on top of it and its diff assumes its commits are present.

---

## Why

Cloning this repo and getting to a running bot took an afternoon: twelve Discord channel IDs pasted in by hand, a full `docker-compose up --build` on every code change, and a fresh clone that didn't compile at all. This fixes that path end to end.

## Reviewing this

140 files, but most of it is mechanical. Suggested order:

| Read | Skip |
|---|---|
| `src/config.ts` — the one real new module | `74ca28b` "apply biome formatting" — pure reformat |
| `scripts/setup.mjs`, `scripts/dev.mjs`, `scripts/postinstall.mjs` | `c871480` "normalise game data to LF" — line endings only, 12k lines |
| `.github/workflows/docker-build.yml` — the new smoke job | "biome lint autofixes" — mostly `import type` |
| `Dockerfile`, `docker-entrypoint.sh` | |
| `README.md`, `CONTRIBUTING.md` | |

Both reformat commits are in `.git-blame-ignore-revs`. Excluding the line-ending commit, the real diff is **137 files, +6,070 / −7,228**.

Commits are organised by theme and each one stands alone, so reviewing commit-by-commit works better than the combined diff.

---

### Fresh clone now builds

`src/prisma/generated` is gitignored and only `npm run build` ran `prisma generate` — so `npm install && npm run dev` dropped you into **72 TypeScript errors** that look like a broken repo. A guarded `postinstall` generates the client; `dev.mjs` does too, so the loop is self-sufficient.

The guard matters: at `npm ci` time the Dockerfile has only copied `package*.json`, so neither `scripts/` nor `.husky/` exists yet. The image build is now `npm ci --ignore-scripts`, which is correct regardless — git hooks are meaningless in an image and `npm run build` runs `prisma generate` explicitly a few lines later.

### Setup: twelve channel IDs → one command

```sh
npm run setup -- <guild-id>          # creates the 9 channels, writes their ids to .env
npm run setup -- <guild-id> --dry-run
```

Shows a plan and asks before touching your server, adopts existing channels by name rather than duplicating, and is safe to re-run after a partial failure. REST only — no gateway connection, no command registration — so it can't disturb a live bot. The server list is parsed from `schema.prisma` rather than duplicated.

### Config validated at startup

New `src/config.ts` validates every environment variable on first read and reports **all** problems at once rather than one per restart. An empty environment lists all 15 missing vars with a pointer to `npm run setup`.

Previously a missing variable surfaced as `undefined` deep inside whatever first needed it — a thumbnail URL built as `undefinedItem_Name.png`, a channel fetch against the id `undefined`, a watch duration of `NaN` days. Silent for a while, then wrong.

`npm run doctor` runs the same check without connecting to Discord, Postgres, or Redis.

Per-server variables derive from the Prisma `Server` enum, so adding a server to the schema adds its required variables rather than failing the first time that server sees an auction.

### Dev loop: rebuild → ~1s reload

```sh
npm run dev:deps   # postgres + redis in docker
npm run dev        # the bot on your host, reloading on save
```

`docker-compose.dev.yml` layers over the production compose file to publish 5432/6379 on localhost; the production stack keeps its unix sockets and is untouched. The dev script supplies `DATABASE_URL`, `REDIS_URL`, and `FAKE_LOGS`, so **`.env` only needs the Discord section and no EverQuest client is required to develop.**

### CI: proves the container can start, not just that it builds

Building the image proves it compiles. It doesn't prove it runs — the Dockerfile copies runtime files through a hand-maintained list of paths, and `docker-entrypoint.sh` is 77 lines of migration and retry logic that nothing executed.

New `smoke` job brings up the real compose stack and runs the real image in `SMOKE_TEST` mode: apply migrations, validate config and runtime assets, exit. Verified locally that deleting the `schema.prisma` COPY still builds a green image and fails this job with exit 1.

Also: `typecheck` added to CI; `whoan/docker-build-with-cache-action` (third-party) replaced with `docker/build-push-action` + GHA cache; concurrency group so superseded PR runs cancel.

### Tooling: ESLint + Prettier → Biome

**10 packages → 1** (devDeps 20 → 10). Lint drops from ~3.3s to ~70ms.

Two packages the old config actually used — `typescript-eslint` and `@eslint/js` — were never declared; they resolved through `eslint-config-love`, which the config no longer referenced. Removing that unused package would have broken linting.

The old setup also paid for type-aware linting without using it: `parserOptions.project: true` built the full TS program on every run while the config used `configs.recommended`, not `recommendedTypeChecked`. No type-aware rule was enabled, so nothing is lost.

Test files were excluded from linting entirely (`**/*.test.ts`, `src/test/**`, `test/**`). Biome covers them.

### Windows contributors

`.gitattributes` now normalises text to LF. Biome enforces `lineEnding: "lf"` and this project's contributors run EverQuest, i.e. Windows — without this, a default `core.autocrlf=true` checkout fails lint on every file before anything is edited. It already happened once, in the branch below this one.

`src/exampleLogs/` is exempt: those are captured EverQuest logs and the game writes CRLF, so normalising them would make the sample less faithful than the real input the parser tails.

### Open-source scaffolding

LICENSE (package.json has claimed ISC since 2022 with no file), CONTRIBUTING.md with an add-a-command walkthrough, issue/PR templates, `dependabot.yml` for scheduled version bumps.

The bug template asks for the auction line verbatim — for parser issues that single line is usually enough to write a failing test from.

### Removed the `debug_console` global

`globalThis.debug_console` was assigned inside `startLoggingAllServers()` in `lib/parser` but called from twenty button handlers in `lib/content`. It worked purely on startup ordering. Replaced with `src/lib/helpers/logger.ts`; `src/global.d.ts` and the test stub are gone.

---

## New scripts

| | |
|---|---|
| `npm run check` | lint + typecheck + tests — everything CI gates on |
| `npm run doctor` | validate `.env`, no connections made |
| `npm run setup -- <guild-id>` | provision Discord channels |
| `npm run dev` / `dev:deps` | host dev loop |
| `npm run typecheck` | `tsc --noEmit` |

Removed `fakeLogs` (folded into `dev`). `dev`, `lint`, `lint:fix`, `format` changed.

## Dependencies

- **devDependencies 20 → 10** — dropped `eslint` + 8 plugins/configs, `prettier`, `concurrently` (unused since `dev.mjs` spawns children directly); added `@biomejs/biome`
- **dependencies 13 → 14** — added `zod` for config validation

## Verification

- **575 unit tests** (up from 532) + **55 integration tests**, lint and typecheck clean
- Fresh-clone fix verified by deleting `src/prisma/generated`: 72 errors → 0
- Docker image builds; compose smoke test exits 0
- `npm run dev` verified end to end against a test guild with a deliberately invalid token, so it never touched the live bot's command registration

## Notes for the merger

- Branch history was rebased — push needs `--force-with-lease`
- Two commits are in `.git-blame-ignore-revs`; GitHub honours it automatically, locally it's `git config blame.ignoreRevsFile .git-blame-ignore-revs`
- Nothing here has run on GitHub Actions yet — the smoke job, concurrency group, and dependabot config are verified locally only


jamesjamail:
# Developer experience: setup, tooling, and CI

Makes this repo something a stranger can clone and contribute to. No runtime behaviour changes to the bot itself.

> **Stacked on #<ai-fixes-review PR>.** Merge that one first — this branch is rebased on top of it and its diff assumes those three commits are present.

---

## Why

Cloning this repo and getting to a running bot took an afternoon: twelve Discord channel IDs pasted in by hand, a full `docker-compose up --build` on every code change, and a fresh clone that didn't compile at all. This fixes that path end to end.

## Reviewing this

141 files, but most of it is mechanical. Suggested order:

| Read | Skip |
|---|---|
| `src/config.ts` — the one real new module | `196e4b7` "apply biome formatting" — pure reformat |
| `scripts/setup.mjs`, `scripts/dev.mjs`, `scripts/postinstall.mjs` | `0f0c786` "normalise game data to LF" — line endings only, 12k lines |
| `.github/workflows/docker-build.yml` — the new smoke job | `8ff42d6` "biome lint autofixes" — mostly `import type` |
| `Dockerfile`, `docker-entrypoint.sh` | |
| `README.md`, `CONTRIBUTING.md` | |

Both reformat commits are in `.git-blame-ignore-revs`. Excluding the line-ending commit, the real diff is **138 files, +6,419 / −7,506**.

Commits are organised by theme and each one stands alone, so reviewing commit-by-commit works better than the combined diff.

---

## What changed

### Fresh clone now builds

`src/prisma/generated` is gitignored and only `npm run build` ran `prisma generate` — so `npm install && npm run dev` dropped you into **72 TypeScript errors** that look like a broken repo. A guarded `postinstall` generates the client; `dev.mjs` does too, so the loop is self-sufficient.

The guard matters: at `npm ci` time the Dockerfile has only copied `package*.json`, so neither `scripts/` nor `.husky/` exists yet. The image build is now `npm ci --ignore-scripts`, which is correct regardless — git hooks are meaningless in an image and `npm run build` runs `prisma generate` explicitly a few lines later.

### Setup: twelve channel IDs → one command

```sh
npm run setup -- <guild-id>          # creates the 9 channels, writes their ids to .env
npm run setup -- <guild-id> --dry-run
```

Shows a plan and asks before touching your server, adopts existing channels by name rather than duplicating, and is safe to re-run after a partial failure. REST only — no gateway connection, no command registration — so it can't disturb a live bot. The server list is parsed from `schema.prisma` rather than duplicated.

### Config validated at startup

New `src/config.ts` validates every environment variable on first read and reports **all** problems at once rather than one per restart. An empty environment lists all 15 missing vars with a pointer to `npm run setup`.

Previously a missing variable surfaced as `undefined` deep inside whatever first needed it — a thumbnail URL built as `undefinedItem_Name.png`, a channel fetch against the id `undefined`, a watch duration of `NaN` days. Silent for a while, then wrong.

`npm run doctor` runs the same check without connecting to Discord, Postgres, or Redis.

Per-server variables derive from the Prisma `Server` enum, so adding a server to the schema adds its required variables rather than failing the first time that server sees an auction.

### Dev loop: rebuild → ~1s reload

```sh
npm run dev:deps   # postgres + redis in docker
npm run dev        # the bot on your host, reloading on save
```

`docker-compose.dev.yml` layers over the production compose file to publish 5432/6379 on localhost; the production stack keeps its unix sockets and is untouched. The dev script supplies `DATABASE_URL`, `REDIS_URL`, and `FAKE_LOGS`, so **`.env` only needs the Discord section and no EverQuest client is required to develop.**

### CI: proves the container can start, not just that it builds

Building the image proves it compiles. It doesn't prove it runs — the Dockerfile copies runtime files through a hand-maintained list of paths, and `docker-entrypoint.sh` is 77 lines of migration and retry logic that nothing executed.

New `smoke` job brings up the real compose stack and runs the real image in `SMOKE_TEST` mode: apply migrations, validate config and runtime assets, exit. Verified locally that deleting the `schema.prisma` COPY still builds a green image and fails this job with exit 1.

Also: `typecheck` added to CI; `whoan/docker-build-with-cache-action` (third-party) replaced with `docker/build-push-action` + GHA cache; concurrency group so superseded PR runs cancel.

### Tooling: ESLint + Prettier → Biome

**10 packages → 1** (devDeps 20 → 10). Lint drops from ~3.3s to ~70ms.

Two packages the old config actually used — `typescript-eslint` and `@eslint/js` — were never declared; they resolved through `eslint-config-love`, which the config no longer referenced. Removing that unused package would have broken linting.

The old setup also paid for type-aware linting without using it: `parserOptions.project: true` built the full TS program on every run while the config used `configs.recommended`, not `recommendedTypeChecked`. No type-aware rule was enabled, so nothing is lost.

Test files were excluded from linting entirely (`**/*.test.ts`, `src/test/**`, `test/**`). Biome covers them.

### Windows contributors

`.gitattributes` now normalises text to LF. Biome enforces `lineEnding: "lf"` and this project's contributors run EverQuest, i.e. Windows — without this, a default `core.autocrlf=true` checkout fails lint on every file before anything is edited. It already happened once, in the branch below this one.

`src/exampleLogs/` is exempt: those are captured EverQuest logs and the game writes CRLF, so normalising them would make the sample less faithful than the real input the parser tails.

### Open-source scaffolding

LICENSE (package.json has claimed ISC since 2022 with no file), CONTRIBUTING.md with an add-a-command walkthrough, issue/PR templates, `dependabot.yml` for scheduled version bumps.

The bug template asks for the auction line verbatim — for parser issues that single line is usually enough to write a failing test from.

### Removed the `debug_console` global

`globalThis.debug_console` was assigned inside `startLoggingAllServers()` in `lib/parser` but called from twenty button handlers in `lib/content`. It worked purely on startup ordering. Replaced with `src/lib/helpers/logger.ts`; `src/global.d.ts` and the test stub are gone.

---

## New scripts

| | |
|---|---|
| `npm run check` | lint + typecheck + tests — everything CI gates on |
| `npm run doctor` | validate `.env`, no connections made |
| `npm run setup -- <guild-id>` | provision Discord channels |
| `npm run dev` / `dev:deps` | host dev loop |
| `npm run typecheck` | `tsc --noEmit` |

Removed `fakeLogs` (folded into `dev`). `dev`, `lint`, `lint:fix`, `format` changed.

## Dependencies

- **devDependencies 20 → 10** — dropped `eslint` + 8 plugins/configs, `prettier`, `concurrently` (unused since `dev.mjs` spawns children directly); added `@biomejs/biome`
- **dependencies 13 → 14** — added `zod` for config validation

## Verification

- **575 unit tests** (up from 531) + **55 integration tests**, lint and typecheck clean
- Fresh-clone fix verified by deleting `src/prisma/generated`: 72 errors → 0
- Docker image builds; compose smoke test exits 0
- `npm run dev` verified end to end against a test guild with a deliberately invalid token, so it never touched the live bot's command registration

## Notes for the merger

- Branch history was rebased — push needs `--force-with-lease`
- Two commits are in `.git-blame-ignore-revs`; GitHub honours it automatically, locally it's `git config blame.ignoreRevsFile .git-blame-ignore-revs`
- Nothing here has run on GitHub Actions yet — the smoke job, concurrency group, and dependabot config are verified locally only

rm-you (commented):
Reviewed commit-by-commit as suggested. The reading order in the description helped — thanks for writing it.

Overall this is a real improvement and the mechanical/substantive split is honest. Ignoring line-ending and format churn the diff is ~2,900 added / ~650 removed, which is reviewable. Findings below, ordered by how much I think they matter.

## First-run path is broken in two places

These both sit on the path a new contributor takes, which is the path this PR exists to fix.

**`npm run doctor` cannot run on a fresh clone.** `package.json` has `"doctor": "node ./build/doctor.js"`, and nothing in the quickstart builds — `postinstall` runs `prisma generate`, not `tsc`. So the sequence in the README gets `MODULE_NOT_FOUND`. This is also the last thing `setup.mjs` prints (`Run \`npm run doctor\` to confirm the configuration is complete`) and what `.env.example`'s header recommends "whenever something will not start". Either have the script build first, or run it from source.

**`.env.example` ships `FAKE_LOGS=false`, so `doctor` fails on a correct file.** The `SERVERS_*_LOG_FILE_PATH` variables are composed by compose from the `_LOG_FILE` names, so nothing is misnamed and the design is fine — but on the host they don't exist, `parseConfig` throws, and `doctor` exits 1. Your hint text explains this well, which is what stops it being confusing, but a non-zero exit still reads as "your config is broken" on a config that's correct for host development. Given the file's own header says host dev only needs the Discord section, `FAKE_LOGS=true` looks like the friendlier default. Alternatively treat the missing-log-path case as a warning when the `_PATH` variables are absent entirely.

## `dev.mjs`

**The env precedence comment is backwards.** Lines 20-22 say "an explicit value in .env still wins over these and both still lose to whatever is already exported." But the defaults go into `process.env` before the child is spawned (32-34, `env: process.env` at 43), and dotenv won't override an existing entry — so the dev default beats `.env`. Practical effect: a contributor who has real EverQuest logs and sets `FAKE_LOGS=false` in `.env` silently still gets generated auctions, and a `DATABASE_URL` pointing anywhere but localhost:5432 is ignored under `npm run dev`. Only an exported shell variable actually wins.

The behaviour may well be what you want for a dev script. The comment asserting the opposite is the problem, because it's the thing someone will trust when their override appears not to work.

**The 4-second delay can take down the whole loop.** `logFaker` is spawned at line 100 without `--watch`, so if the first `tsc` pass hasn't emitted `build/lib/parser/logFaker.js` within 4s — cold clone with no `.tsbuildinfo`, Windows AV scanning several hundred emitted files — it exits non-zero, and the handler at 46-52 tears down everything with `[dev] logFaker exited (1); stopping.` That lands on exactly the person the script is for, on their first run, and the message doesn't suggest "try again". Polling for the file, or `--watch` on `logFaker` too, removes the race rather than widening it.

**Children aren't reliably reaped on Windows.** `shell: process.platform === 'win32'` (42) means each child is a `cmd.exe` wrapper; `child.kill('SIGTERM')` (62) signals the wrapper, and `shutdown` calls `process.exit(code)` without waiting, so the `tsc`/`node` grandchildren can survive. Repeated Ctrl-C leaves watchers holding `build/`. Worth caring about here specifically because this project's contributors are on Windows by definition.

## `setup.mjs`

The keep/adopt/create split is the right shape, and `parseServerNames` reading the enum out of `schema.prisma` — textually, so it works before `prisma generate` — is my favourite thing in this PR. Three things I'd change:

**`.env` is rewritten without asking in two paths.** The prompt gates on `creating.length > 0` (316), but `writes` includes adopts (303) and the write at 349 is unconditional. So (a) if all nine channels already exist by name, `.env` is rewritten with no confirmation, and (b) `--force` replaces every configured id with no confirmation. The plan is printed first, so it isn't silent — but `--force` overwriting ids someone may have hand-picked is the single most regrettable mutation here and it's the one that doesn't ask.

**The error being diagnosed is discarded.** At 277-286 only `error.status` is used, never `error.message`, and `fail()` exits before `main().catch` at 362 could print it. A DNS failure, corporate proxy, TLS interception, or a plain `TypeError` from a future edit all render as `HTTP ?` followed by three bullets about guild ids — advice for a problem the user doesn't have. Everything else in this script has unusually good error messages, which is what makes this one stand out.

**`applyEnvUpdates` updates the first occurrence of a key; every reader takes the last.** 129-132 does `remaining.delete(key)` after the first match, while `parseEnvFile` (152-159) and dotenv both let the last occurrence win. If a `.env` has a key twice, setup rewrites the line nobody reads and the stale value keeps winning, with a success message. Rare, but silent when it happens.

Smaller: `confirm()` has no `process.stdin.isTTY` check, so the prompt can't be answered under a pipe or in CI; `writeFileSync('.env', ...)` is non-atomic on the only copy of the bot token, where a temp-file-and-rename is cheap; and both `readFileSync` calls are cwd-relative, so running the script from a subdirectory reports "No .env found. Copy .env.example to .env first."

## Smoke job

Removing the `gameData` COPY *and* adding the `doctor` assertion that proves it was redundant, in the same PR, is the right way to delete something you believe is dead. I checked the chain and it holds: static JSON import plus `resolveJsonModule` means tsc emits it, and the `itemCount === 0` check catches it if that ever stops being true. The migration path is genuinely exercised too — `prisma.config.ts`, `schema.prisma`, and `migrations/` are all load-bearing for that step.

But "proves the container can start" is broader than what runs. `doctor.ts` imports config, game data and the Prisma client — it never loads `registerCommandHandlers`/`registerEventHandlers`, which `readdirSync` the build tree and `require()` each file at boot. That runtime filesystem scan is the largest "builds clean, dies at boot" surface in this repo and smoke doesn't touch it. Two smaller gaps: `src/redis/init` is never imported even though the Redis container is up, and because `SMOKE_TEST` short-circuits before the `FAKE_LOGS` branch, `logFaker.js` is never exercised despite the fixture setting `FAKE_LOGS=true`. Having `doctor` load the command and event directories would close most of this — neither needs the network, only `registerSlashCommands` does.

Related: `is_retryable_failure` classifies `*ENOENT*` as retryable (entrypoint:20). A missing `schema.prisma` or `migrations/` — precisely what smoke is meant to catch — would retry 30 times at 2s before failing. Pre-existing, but the smoke job is its new consumer and a fast failure is the whole point.

## Publish job

Push gating is correct twice over (`on.push.branches: [main]` plus `if: github.event_name == 'push'`), and the concurrency expression can't cancel a main publish since `github.ref` for a PR is never `refs/heads/main`. Replacing an unmaintained third-party action with `docker/build-push-action` is a straight win.

Two things: the image is built twice per main push — `docker compose build` in smoke with no buildx or GHA cache, then the real build in `build-image-and-push` — so the artifact that gets published isn't the one that passed smoke. Build-and-push by digest, smoke that digest, then tag would fix both. And only `:latest` is tagged (129), so there's no rollback target and no way to pin a deploy; adding `:${{ github.sha }}` to `tags` is one line. Also `permissions: packages: write` is workflow-wide (8-10), so `test`, `integration` and `smoke` all get registry write on same-repo PR runs — moving it into the publish job costs nothing.

## `config.ts`

The module earns its place — I checked all the consumers, and rewriting `getEnvironmentVariable` to read through `config()` is what closes the escape hatch rather than adding a second way to do things. Lazy-and-memoised is the right shape, and reporting every problem at once is the part that actually saves time.

**The `Config` type is hand-maintained and applied with an unchecked cast** (99-114, `result.data as Config` at 162). Change `WATCH_DURATION_IN_DAYS` in the schema and the type keeps claiming `number` with no error. `z.infer<ReturnType<typeof buildSchema>>` would derive it. The `& Record<string, unknown>` is the sharper half: it makes `config().TYPOED_KEY` compile and return `undefined`, which is the exact failure mode the module's own header comment says it exists to eliminate. The dynamic server keys do need an index signature, but scoping it to a template-literal type would keep typo detection on the fixed keys.

**`WATCH_DURATION_IN_DAYS=` with a blank value becomes 0, not 7.** `.default()` only fires on `undefined`, so an empty string coerces to 0 and fails `.positive()` — reported as "expected number to be >0" rather than falling back. A key present but blanked is a normal thing to find in a `.env`. Preprocessing `''` to `undefined` fixes it.

**`DATABASE_URL: z.string().min(1)`** (74) is the weakest check on the one URL that's assembled from parts. The other three get `z.url()`. If a `POSTGRES_*` variable is missing, expansion yields `postgresql://:@localhost/?host=` — non-empty, passes, and resurfaces later as a connection error, which is the deferred-failure pattern this module is meant to end.

## Logger

Clean conversion — 21 call sites, and `git grep debug_console` comes back with only the explanatory comment. Keeping the `/^[tT]/` spelling deliberately, with a test named for the invariant, is the right way to do a behaviour-preserving refactor.

Worth noting in the description that this fixes a real bug rather than just tidying: on main the global was assigned inside `startLoggingAllServers()`, which runs after `initializePrisma()`, and `ready.ts` retries that indefinitely. So while Postgres was unreachable the bot was live and accepting interactions with `debug_console` unassigned — and in these handlers the call is the last statement, after `interaction.update()` had already fired. Users saw the button work; the log showed a `ReferenceError`. That's now structurally impossible.

One new coupling: `debug()` reads `config()`, which throws `ConfigError` on an incomplete environment, so a log statement can now fail where the old global couldn't. Unreachable today — `index.ts` validates and exits at boot, and `src/test/env.ts` supplies a full environment — but it's 19 interaction handlers' last statement, so it's worth either reading `process.env.DEBUG_MODE` directly in the logger or swallowing the throw.

Minor: `src/test/env.ts:27` hardcodes `['BLUE', 'GREEN', 'RED']` while `config.ts:16` derives `SERVER_NAMES` from the enum. Adding a server to `schema.prisma` therefore breaks every unit test that reaches `config()`. Your comment argues that's acceptable because it breaks loudly, which is fair — but importing the enum would give you the property the module advertises.

## Biome

Compared the configs directly; this holds up better than tooling swaps usually do. Coverage genuinely widens — the old config ignored `**/*.test.ts`, `**/*.itest.ts`, `src/test/**`, `test/**` and the root config files, so a meaningful fraction of the repo was unlinted. Only two rules are relaxed (`noNonNullAssertion`, `useNodejsImportProtocol`), both stylistic, and `--error-on-warnings` keeps `noExplicitAny` enforced despite being `warn`. The undeclared-transitive-dependency point about `typescript-eslint` and `@eslint/js` is a good catch and worth having found regardless of the swap.

One thing I'd drop: several cases in `test/config.test.ts` assert on source text rather than behaviour — `expect(indexSource).toMatch(/process\.exit\(1\)/)` passes for any `process.exit(1)` anywhere in the file, and the check for the literal `FAKE_LOGS: 'true'` in `dev.mjs` breaks on a quote-style change while behaviour is identical. They'll produce false failures and miss real regressions. The blame-ignore and biome-JSON checks are the opposite — both guard things that fail silently — and I'd keep those.

Also: `tsconfig.json` only includes `./src/**/*`, so `npm run typecheck` never sees the new `test/` directory. That makes the `@ts-expect-error` in `test/setup.test.ts` inert, since an unused one would normally be an error itself.

---

Nothing here is an objection to the direction. The first-run items are the ones I'd fix before merge, since a contributor following the README hits them immediately; most of the rest is comment-accuracy and least-privilege tightening.

rm-you:
One follow-up I left out of the review above, because it connects two things I raised separately and I wanted to confirm the chain before asserting it.

`src/lib/helpers/errors.ts` imports the client from the entrypoint:

```ts
import { client } from '../..';
```

`index.ts` reaches `errors.ts` through its own graph, so `index -> errors -> index` is a cycle. That's pre-existing — byte-identical on `main` — and it has always been harmless, because CommonJS binds `client` late and by the time an error is reported the export is populated. I wouldn't raise it on its own.

This PR changes the consequence, though. `index.ts` now calls `appConfig()` at module scope and `process.exit(1)` on `ConfigError`. Combined with the hardcoded server list in `src/test/env.ts:27` and the fact that `src/test/setup.ts` installs no global mock for the entrypoint — each test file does `vi.mock('../../index', ...)` individually, ~40 of them — the chain is:

1. someone adds a server to `schema.prisma`
2. `config()` derives the requirement from the enum (`config.ts:16`) and now demands two more variables
3. `src/test/env.ts` doesn't supply them, so `config()` throws `ConfigError`
4. any test that reaches `errors.ts` without mocking `../../index` evaluates the entrypoint's module scope
5. `process.exit(1)` kills the vitest worker

Step 5 is the problem. A killed worker reports as a crashed suite with no failing assertion, which is a much worse first day than "these two variables are missing." And the trigger in step 1 is exactly the change the config module advertises as safe to make.

Since you're already in this area, `src/client.ts` would be worth doing here rather than later:

```ts
// src/client.ts
export const client = new Client({ intents: [...], partials: [...] });
```

`index.ts` imports it and logs in; `errors.ts` and the other consumers import from `src/client` instead of `../..`. That breaks the cycle, makes the dependency honest — these modules need the client, not the entrypoint — and removes the reason for the per-file `vi.mock('../../index')` boilerplate, which is currently the only thing standing between an incomplete test environment and a dead worker.

If you'd rather not grow this PR, the cheap mitigations are to have `src/test/env.ts` derive its server list from the `Server` enum the way `config.ts` does, or to move the entrypoint mock into `src/test/setup.ts` so it can't be forgotten. Either one breaks the chain at a different link. The extraction is the one that stops it being a chain at all.

rm-you:
Two notes on the last two commits, neither blocking.

**On dropping the blame-ignore guard test — you were right and I was wrong.** I said above to keep it, on the grounds that it guarded something that fails silently. Your commit message points out it was itself failing silently: `git cat-file -e` only checks that the object exists, so a local backup tag keeps an orphaned commit reachable and the test passes against a file that has already gone stale. `git merge-base --is-ancestor` was the check it wanted.

The squash-merge argument for deleting the file outright is the stronger point anyway, and #97 collapsing into `12b051c` is the demonstration — none of its commit subjects survive on main, so there was never going to be a reformat commit there to skip.

**On `engine-strict`:** it reaches further than the root `engines` field — npm's docs have it refusing to install "any package that claims to not be compatible with the current Node.js version", so dependencies are in scope too. Nothing in the tree trips it today, but a future dep declaring something like `^18 || ^20` becomes a hard `npm ci` failure rather than an EBADENGINE warning, with `--force` or an override as the only way out. Still worth having; just a wider blast radius than the commit message implies.

jamesjamail:
Superseded — split into two PRs, as @rm-you suggested:

- #106 — mechanical only: biome adoption, reformat, line-ending normalisation. Merge this first.
- #107 — the substantive half, stacked on #106.

Everything here is carried over; I verified no file and no content is lost against the pair.

#107 also contains the follow-ups to the review on this PR — the first-run `doctor` and `FAKE_LOGS` bugs, the `dev.mjs` env precedence and startup race, the `setup.mjs` confirmation and error handling, the derived `Config` type, the logger coupling, `ENOENT` retry classification, and the CI permission and tagging changes. Two items are called out as deliberately not addressed at the bottom of its description.

Thanks for the commit-by-commit read — the split came directly out of it.

PR #107, opened 2026-09-04
DX improvements: config validation, host dev loop, channel setup, smoke test

jamesjamail:
The substantive half of #103, stacked on #106. **Merge #106 first.**

Ignoring the mechanical churn now in #106, this is 88 files, +2,420 / −513 — close to the reviewable core @rm-you measured in his review of #103.

Everything he raised there is addressed except two items called out at the bottom.

## Reading order

1. `chore: pin node, add typecheck script, build image from lockfile`
2. `fix: resolve lint findings biome surfaced` — the 13 pre-existing findings, plus the CI lint gate deferred by #106
3. `docs: update linter section for biome`
4. `feat: validate configuration at startup` — `src/config.ts`, `src/doctor.ts`
5. `feat: run the bot on the host with a fast reload loop` — `scripts/dev.mjs`
6. `ci: prove the container can start, not just that it builds` — the smoke job
7. `feat: provision discord channels with npm run setup`
8. `refactor: reconcile the env helper with startup config validation`
9. `fix(test): give the integration suite a complete environment`
10. `fix: make a fresh clone build, and stop CRLF reaching the repo`
11. `docs: add the open-source scaffolding this repo was missing`
12. `refactor: replace the debug_console global with a logger module`
13. `chore: fail the install on the wrong Node version`
14. `fix: make the first-run path work from a fresh clone`
15. `ci: re-enable the lint gate now that the findings are fixed`
16. — 21. review follow-ups, below

## First-run fixes

`npm run doctor` mapped to `node ./build/doctor.js` while nothing in the quickstart compiled, so a fresh clone got `MODULE_NOT_FOUND` — from the command `setup.mjs` prints on success and `.env.example` recommends. It builds first now.

`.env.example` shipped `FAKE_LOGS=false`, so `doctor` exited 1 on a correct file. The shipped `LOG_SOURCE_PATH` is a Windows placeholder, so `false` was never a working default.

Both verified from a pristine `.env.example` with only the Discord section filled.

## Review follow-ups

**A module import no longer boots the bot.** `index.ts` called `login()` and `registerSlashCommands()` at module scope, and six modules import `client` from it — `errors.ts` among them, which most of the codebase reaches. Importing almost anything opened a gateway connection and re-registered slash commands against the live application. Startup is now behind `require.main === module`.

That is what lets `doctor` cover the boot-time `readdirSync` + `require()` of both handler directories — the largest "builds clean, dies at boot" surface here, and the gap in the smoke job. Verified offline: zero `login()` calls, zero REST calls. Also imports `src/redis/init`, which nothing was exercising.

**`dev.mjs` env precedence was backwards.** The comment promised `.env` beat the script defaults; the defaults went into `process.env` first and dotenv will not overwrite, so they won silently. Someone with real logs setting `FAKE_LOGS=false` still got generated auctions. The behaviour now matches the promise. The fixed 4s delay before spawning children is replaced by polling, and Windows shutdown uses `taskkill /T`.

**`setup.mjs` asks before any write.** The prompt gated on there being channels to create, so adopt-only runs and `--force` both rewrote `.env` unasked. Non-HTTP errors are re-thrown rather than rendered as guild-id advice. `applyEnvUpdates` rewrites the *last* occurrence of a duplicated key, which is the one that wins. Plus a TTY check, an atomic write, and repo-relative paths.

**`config.ts` types are derived.** `Config` is `z.infer` plus a template-literal union for the server keys, replacing a hand-maintained type and `& Record<string, unknown>` — `config().TYPOED_KEY` is a compile error again. Blank `WATCH_DURATION_IN_DAYS` falls back to 7. `DATABASE_URL` has to actually resolve.

**A log line can no longer throw.** `debug()` read `config()`, which throws on an incomplete environment, and it is the last statement in 19 interaction handlers.

**Entrypoint fails fast on `ENOENT`**, which is what the smoke job exists to catch. Tested by executing the real shell function.

**CI**: registry write scoped to the publish job; images tagged with the commit sha as well as `latest`.

**Tests**: `typecheck` now covers `test/`, which immediately caught two `@ts-expect-error` directives that were inert. Whole-file `process.exit(1)` matches now match handler bodies.

## Verification

596 tests, lint, typecheck green. Smoke verified against the real compose stack: baseline passes; removing the events directory or emptying the commands directory each fails it. Every fix has a test confirmed to fail without it.

## Not addressed

**logFaker is still not exercised by smoke** — `SMOKE_TEST` short-circuits before the `FAKE_LOGS` branch, and reordering that risks the migration sequencing the tests pin.

**The image is still built twice**, so the published artifact is not the one smoke tested. Fixing it properly means build-and-push by digest, smoke that digest, then tag — a rewrite of the job graph. This is the most valuable thing left.

Both are worth their own PR.


CONTRIBUTING.md
```diff
@@ -0,0 +1,126 @@
...
+
+You need Docker and Node 24 (`.nvmrc` pins it). You do **not** need EverQuest —
+the dev loop generates fake auction lines.
+
+```sh
+npm install
+cp .env.example .env      # then fill in the Discord section
+npm run setup -- <guild-id>   # creates the channels, writes their ids into .env
+npm run dev:deps          # postgres + redis in docker
+npm run dev               # the bot, on your host, reloading on save
```
ChuckHend: As far as I can tell, the SQL migrations will not have run up to this point so `npm run dev` will error out complaining about tables missing in Postgres. The migrations do get applied in [docker-entrypoint.sh](https://github.com/jamesjamail/TunnelQuestBot/blob/8a540bd622951d78135541ad30bda3d4e95c8ad3/docker-entrypoint.sh#L69), but these instructions are going around that.

So there could just be another step in these instructions after `npm run dev:deps`, `DATABASE_URL=postgres://.... npm run migrate`. Could also look at baking it into dev.mjs somewhere.

scripts/dev.mjs
```diff
@@ -0,0 +1,245 @@
...
+			shell: process.platform === 'win32',
+		});
+	} catch {
+		console.error(
+			'[dev] prisma generate failed — run `npm run generate` to see why',
+		);
+		process.exit(1);
+	}
+
+	const fakeLogs = process.env.FAKE_LOGS === 'true';
```
rm-you: **The shipped `.env` enables fake logs, but this evaluates to false in the launcher.**

`resolveDevEnv()` skips defaults for keys present in `.env`, and `main()` only assigns those returned defaults into `process.env`. It never loads the `.env` values themselves into the launcher.

With `.env.example` copied to `.env` and no exported `FAKE_LOGS`:

1. `.env` contains `FAKE_LOGS=true`.
2. The helper therefore omits `FAKE_LOGS`.
3. This expression sees `undefined` and evaluates to false.
4. The launcher skips spawning logFaker.
5. The bot subsequently loads `.env`, sees true, and expects generated log files.

I exercised the launcher with mocked subprocesses: it printed “FAKE_LOGS is off” and spawned only tsc and the bot.

Please resolve the effective environment before making launcher decisions, and use those same values for the children. A regression test should use `.env` containing `FAKE_LOGS=true` and assert that logFaker is spawned—not just that the helper omits a default.

Also, use the same boolean interpretation as the bot: its existing parser accepts values such as `True`, while this exact-string comparison does not.

scripts/dev.mjs
```diff
@@ -0,0 +1,245 @@
...
+ * @param {Record<string, string>} defaults
+ * @param {Record<string, string | undefined>} env
+ * @param {Record<string, string>} fromEnvFile
+ * @returns {Record<string, string>}
+ */
+export function resolveDevEnv(defaults, env, fromEnvFile) {
+	/** @type {Record<string, string>} */
+	const applied = {};
+	for (const [key, value] of Object.entries(defaults)) {
+		if (key in env || key in fromEnvFile) continue;
```
rm-you: **The precedence fix needs a corresponding change to the shipped database configuration.**

`.env.example:95` already defines `DATABASE_URL`, so this condition suppresses the host-development TCP default. The sample URL expands to:

`postgresql://tunnelquestbot:tunn3lb0tp4ss@localhost/tunnelquestbot?host=/dbsocket`

`/dbsocket` is the container stack’s socket directory; it is not made available on the host by publishing Postgres on port 5432.

Consequently, the documented “copy `.env.example`, fill in Discord settings, run `dev:deps` and `dev`” workflow selects the container socket instead of the host TCP connection. The example still explicitly says `npm run dev` overrides this URL.

This remains a problem even after loading `.env` into the launcher correctly. Please distinguish the host-development connection from the container URL while preserving genuine user overrides, and add a regression test using the actual `.env.example` as input.

src/index.ts
```diff
@@ -14,16 +14,31 @@ export const client = new Client({
...
+// 	Validate the whole environment before doing anything else. Every consumer
+// 	reads through config(), which memoises this first call, so a bad .env fails
+// 	here with the full list of problems rather than surfacing later as an
+// 	`undefined` interpolated into a URL or a channel id.
+try {
+	appConfig();
+} catch (error) {
+	if (error instanceof ConfigError) {
+		console.error(error.message);
+		process.exit(1);
```
rm-you: **The main-module guard fixes login/registration side effects, but imports can still terminate the process here.**

This configuration validation remains at module scope, outside `boot()` and the `require.main === module` guard. Any consumer reaching `errors.ts → index.ts` with an incomplete environment can still trigger `process.exit(1)` just by importing it.

I evaluated the entrypoint with `require.main !== module` and an injected `ConfigError`: it still called `process.exit(1)`. With valid configuration, it still installed the global rejection/exception handlers.

Deriving the test server list from Prisma fixes the specific “add a server” trigger from my previous comment, which is good. It does not remove the underlying import hazard.

Please move startup-only validation and process handlers behind the entrypoint guard, or extract the client into an import-safe module. A regression test should import a consumer with incomplete configuration and verify that it neither exits nor installs startup handlers.

scripts/setup.mjs
```diff
@@ -0,0 +1,446 @@
...
+	}
+
+	const updates = Object.fromEntries(
+		writes.map((entry) => [entry.envKey, entry.id]),
+	);
+	//	Write through a temp file in the same directory and rename over the
+	//	original. .env holds the only copy of the bot token, and a crash midway
+	//	through a plain write truncates it.
+	const tmpPath = `${envPath}.setup-${process.pid}`;
+	writeFileSync(tmpPath, applyEnvUpdates(envSource, updates));
```
rm-you: **The atomic replacement can make a previously private `.env` readable by other local users.**

The temporary file is created with the default mode, then renamed over the original. On POSIX with umask `022`, that creates a `0644` file even when the existing `.env` was `0600`.

I reproduced this write/rename sequence with a synthetic file: permissions changed from `0600` to `0644`. Since `.env` contains the bot token, this is a permission regression.

Please create the temporary file with restrictive permissions—preferably exclusive creation, e.g. `{ mode: 0o600, flag: 'wx' }`—and preserve or deliberately restrict the destination’s permissions. Add a regression test starting with a `0600` file under umask `022`.

src/doctor.ts
```diff
@@ -0,0 +1,144 @@
...
+//	data JSON and the generated Prisma client, both of which the Dockerfile
+//	copies into the runtime image through a hand-maintained list of paths. If
+//	that list falls behind, the image still builds and this is what notices.
+import { consolidatedItemsAndAliases } from './lib/gameData/consolidatedItems';
+import { Server } from './prisma/client';
+//	Imported for the same reason as the game data: the runtime image copies a
+//	hand-maintained list of paths, and this is the module that notices when the
+//	list falls behind. Importing it does not open a connection - ioredis connects
+//	lazily - so this stays a configuration check.
+import './redis/init';
```
rm-you, line 25: **Nonblocking: the explanation above this import is inaccurate.**

`src/redis/init.ts` constructs `new Redis(...)` without `lazyConnect: true`. ioredis automatically initiates connection by default; lazy connection is an explicit option.

I’m not claiming a connection completes before doctor’s synchronous `process.exit()`. However, that immediate exit should not be the basis for describing the import as connection-free.

Please either make initialization explicitly lazy and test the offline guarantee, or narrow the comment. If changing the initializer, preserve the normal bot’s connection behavior.

rm-you (changes requested):
Reviewed at `8a540bd`, following up on my comments from #103.

Thanks for separating the mechanical changes and addressing the earlier review. The setup confirmation/error handling, config typing, logger decoupling, test-environment server derivation, and publish permissions/tagging are meaningful improvements.

I’m requesting changes because the documented fresh-clone development path still has two problems introduced by the environment-precedence fix:

- The launcher doesn’t load `.env` values before deciding whether to spawn logFaker.
- The shipped container `DATABASE_URL` now suppresses the host-development TCP default.

There are also two fixes requested inline: configuration validation can still exit the process during import, and atomic `.env` replacement can broaden file permissions. The Redis note is nonblocking.

Verification was targeted: I exercised the launcher and entrypoint with mocked dependencies and reproduced the permission change with a synthetic file. I did not run a live bot or the full suite locally. The test, integration, and smoke jobs were green for this commit, but they do not establish that these specific cases work.

The two explicitly deferred items—exercising logFaker in smoke and publishing the exact smoke-tested image—are reasonable follow-up work. The findings above are separate from those deferrals.

CONTRIBUTING.md
jamesjamail: Thank you

rm-you (approved):
Rechecked at `855085a`. All five original reproduction cases are addressed: the shipped configuration starts logFaker and selects the host TCP database, imports no longer exit or install startup handlers, private .env permissions survive setup, and the Redis comment now describes the connection behavior accurately.

Independent reproductions passed, along with 94 relevant development, setup, configuration, and import-safety tests. GitHub's test, integration, and smoke checks are green.

Approving with two nonblocking follow-ups tracked separately: custom socket database overrides in #109 and interpolated FAKE_LOGS settings in #110. Neither affects the documented setup with the shipped configuration.

scripts/dev.mjs
jamesjamail: Fixed in `8c02105`. You were right about the mechanism: `resolveDevEnv` returned only the defaults it had applied, and `main()` read everything else from `process.env`, where a value set solely in `.env` never appears.

It now returns `{ effective, applied }`. `effective` is what the run will actually use, and every launcher decision reads from it; `applied` is only this script's own literal defaults, which is all that gets written into `process.env` — `.env` values can carry `${...}` references that dotenv-expand resolves in the child, so copying them across would break them.

On the test: the spawn decision is now an exported `plannedChildren()` rather than something inferred from which defaults were omitted, so the assertions are about logFaker being started. Three cases run against the real `.env.example` — `spawns logFaker for the fake-log setting the file ships`, `does not spawn logFaker when .env turns it off`, and `spawns logFaker for a True spelling the bot would also accept`.

Boolean parity is an exported `isEnabled()` using the bot's `/^[tT]/`, tested over `true/True/T/TRUE` and `false/False/F/''/undefined`.

Also verified end to end on an empty volume, following the documented flow: the launcher reported FAKE_LOGS on, and logFaker wrote `BLUE.log`, `GREEN.log` and `RED.log`.

scripts/dev.mjs
jamesjamail: Fixed in `8c02105`. A configured value carrying a socket parameter no longer counts as a host-development override — `isContainerOnly()` tests for `[?&]host=`, and the host TCP default wins over it. A URL written without one is honoured, as is an exported shell variable, since an explicit export is a deliberate act in a way the shipped sample is not.

Three regression tests use the real `.env.example`: `ignores the container database url in favour of the host default` (which also asserts the shipped value actually matches `[?&]host=`, so it fails if the sample changes shape), `still honours a database url written for host development`, and `still honours an exported container-shaped url`.

Separately, @ChuckHend found that nothing applied migrations on this path at all — `docker-entrypoint.sh` does it for the container stack and the host loop never runs it, so `dev:deps` left an empty database and the bot failed on missing tables. The launcher now applies them, using the resolved URL rather than whatever is in `.env`, which is only correct because of this fix.

src/index.ts
jamesjamail: Fixed in `18547a3`. Both the configuration check and the two process handlers moved inside `boot()`.

The test took two attempts, and the first one proves your point about CI. I originally wrote it as a child process requiring `build/lib/helpers/errors.js` — which passed locally against a build that happened to be lying around, and failed in CI, where the `test` job runs before `npm run build`. So it was both broken and, worse, capable of asserting against stale output. Rewritten in `855085a` to import the TypeScript source dynamically and watch `process.exit` and the listener counts in-process, with `dotenv` stubbed so `index.ts` cannot reload the environment the test deliberately breaks. No build step, and nothing to go stale.

`does not exit the process on an incomplete environment` and `does not install startup process handlers`; both confirmed to fail when startup is moved back to module scope, and confirmed to pass with `build/` deleted in CI's step order.

Your line about green checks not establishing that these specific cases work was exactly right here.

scripts/setup.mjs
jamesjamail: Fixed in `18547a3`. Reproduced your result first — 600 → 644 under umask 022 — then extracted the write as `writeEnvAtomically()`: temp file created `{ mode: 0o600, flag: 'wx' }`, `chmod` to whatever the original carried, rename, and unlink if the rename fails.

Preserving the previous mode rather than forcing 0600, so a file someone deliberately opened up is not silently tightened either.

Three tests under umask 022: `leaves a private file private`, `does not tighten a file the user deliberately opened up`, and `leaves no temp file behind`. The first fails against the old write.

src/doctor.ts
jamesjamail: You're right, and I took the narrower of your two options. The comment now says this is not connection-free: `src/redis/init.ts` constructs `Redis` without `lazyConnect`, so ioredis starts connecting on import, and the synchronous exit below wins that race rather than the import being safe by construction.

Not making the initializer lazy, because that changes how the bot itself connects, and altering production connection behaviour to make a diagnostic's comment true is the wrong trade. The comment now describes what actually happens.

jamesjamail:
Re-review would be welcome when you have time — four of the five findings were things you verified by hand, and two of them were regressions from my own precedence fix. I've replied on each thread with what changed and how it was checked; they're all marked outdated now, so they're collapsed.

Worth calling out one thing from the `src/index.ts` thread: the CI failure you'd have seen on this branch was a bad test of mine, not a bad fix. The import-safety regression test required build output that the `test` job produces *after* running tests, so it passed locally against a stale build and failed in CI. Rewritten in `855085a` to need no build at all. Your note that green checks don't establish these specific cases held up literally.

The two deferred items now have issues so they don't get lost on merge: #120 (exercise logFaker in the smoke job) and #121 (publish the exact image smoke tested). The second is probably better done after #108, which adds `p99-logger-init` and the two collector services to the compose stack and changes what smoke brings up.

Also left a few comments on #108 — two of them are about how the collector path composes with the existing `LOG_SOURCE_PATH` bind rather than anything in the parser, which reads well.

PR #108, opened 2026-09-06
Add headless P99 Green and Blue log collectors

rm-you:
TunnelQuestBot currently requires EverQuest text log files, which means Green and Blue collectors need full game clients. This adds an optional `p99-loggers` Compose profile that runs one headless `p99-logger-client` for each server and shares their JSONL output with the bot.

The parser auto-detects logger JSONL records and sends auction messages and OOC player-link requests through the same paths as native EQ log lines. Both formats share the UUID pattern for link requests. Malformed records, decode errors, and unrelated channels are ignored.

The published amd64 logger image is pinned to `sha-fbb036a` and includes the P99 asset checksum inventory. The Green and Blue configuration templates collect only auction and OOC channels. Real credentials and collected logs are excluded from Git and Docker build contexts; TQB ships no asset inventory.

Log-file initialization runs before the bot in every Compose configuration; only the collectors require the `p99-loggers` profile. The legacy log mount defaults to an ignored `./logs` directory, so headless setups need no EverQuest directory. Existing deployments can keep their configured log directory. Collectors use configurable host UID/GID, read-only root filesystems, dropped capabilities, private configuration mounts, and health checks.

Validation:
- formatting, lint, type checks, and compilation passed
- 619 unit tests passed, including native text and JSONL parser coverage
- disposable Compose startup checks passed with the collector profile both enabled and disabled; a deliberately delayed initializer completed before the bot could read its log files
- verified default/unset LOG_SOURCE_PATH and explicit legacy-directory overrides
- GitHub CI runs coverage, integration tests, the production image build, and Compose smoke startup against merged main

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,6 +40,78 @@ services:
         retries: 12
         start_period: 20s
 
+    p99-logger-init:
+      image: busybox:1.37.0
+      environment:
+        P99_UID: "${P99_UID:-65534}"
+        P99_GID: "${P99_GID:-65534}"
+      command:
+        - sh
+        - -ec
+        - |
+          mkdir -p /data/green /data/blue /data/red
+          touch /data/green/chat.jsonl /data/blue/chat.jsonl /data/red/chat.log
+          chown -R "$$P99_UID:$$P99_GID" /data/green /data/blue /data/red
+          chmod 750 /data/green /data/blue /data/red
+          chmod 640 /data/green/chat.jsonl /data/blue/chat.jsonl /data/red/chat.log
+      volumes:
+        - p99-logger-data:/data
+      restart: "no"
+
+    p99-green-logger:
+      image: "${P99_LOGGER_IMAGE:-ghcr.io/rm-you/p99-logger-client:sha-fbb036a}"
+      profiles: ["p99-loggers"]
+      user: "${P99_UID:-65534}:${P99_GID:-65534}"
+      command: ["/config/config.json"]
+      depends_on:
+        p99-logger-init:
+          condition: service_completed_successfully
+      volumes:
+        - type: bind
+          source: "${P99_GREEN_CONFIG_FILE:-./p99-logger/green.json}"
+          target: /config/config.json
+          read_only: true
+          bind:
+            create_host_path: false
+        - p99-logger-data:/data
+      read_only: true
+      cap_drop: ["ALL"]
+      security_opt: ["no-new-privileges:true"]
+      restart: unless-stopped
+      healthcheck:
+        test: ["CMD", "/p99-logger-client", "health", "/data/green/health.json"]
+        interval: 30s
+        timeout: 5s
+        retries: 3
+        start_period: 60s
+
+    p99-blue-logger:
+      image: "${P99_LOGGER_IMAGE:-ghcr.io/rm-you/p99-logger-client:sha-fbb036a}"
+      profiles: ["p99-loggers"]
+      user: "${P99_UID:-65534}:${P99_GID:-65534}"
+      command: ["/config/config.json"]
+      depends_on:
+        p99-logger-init:
+          condition: service_completed_successfully
+      volumes:
+        - type: bind
+          source: "${P99_BLUE_CONFIG_FILE:-./p99-logger/blue.json}"
+          target: /config/config.json
+          read_only: true
+          bind:
+            create_host_path: false
+        - p99-logger-data:/data
+      read_only: true
+      cap_drop: ["ALL"]
+      security_opt: ["no-new-privileges:true"]
+      restart: unless-stopped
+      healthcheck:
+        test: ["CMD", "/p99-logger-client", "health", "/data/blue/health.json"]
+        interval: 30s
+        timeout: 5s
+        retries: 3
+        start_period: 60s
+
     tunnelquestbot:
       build:
         context: .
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -51,16 +124,20 @@ services:
...
         REDIS_SOCKET_DIR: "${REDIS_SOCKET_DIR}"
         FAKE_LOGS: "${FAKE_LOGS:-false}"
         DEBUG_MODE: "${DEBUG_MODE:-false}"
       volumes:
         - type: bind
           source: "${LOG_SOURCE_PATH}"
           target: /logfiles
+        - type: volume
+          source: p99-logger-data
+          target: /p99-logger
```
jamesjamail, line 140: `tunnelquestbot` mounts this volume and reads its log paths out of it, but nothing orders it after `p99-logger-init` — its `depends_on` still lists only postgres and redis.

Two ways that bites:

- Someone follows the README, uncomments `SERVERS_BLUE_LOG_FILE_PATH=/p99-logger/blue/chat.jsonl`, then runs a plain `docker compose up -d` without `--profile p99-loggers`. `p99-logger-init` never runs, the file is never created, and `new Tail(logFilePath, ...)` throws on the missing path.
- Even with the profile on, the ordering is incidental — the bot happens to wait on the postgres and redis healthchecks, which usually gives busybox enough time. That's a race being won, not a guarantee.

Adding `p99-logger-init: condition: service_completed_successfully` to the bot's `depends_on` fixes the second. The first is harder, since the init service is profile-gated and the bot isn't — possibly worth having the bot tolerate a not-yet-existing log file, or documenting that these paths only work under the profile.

docker-compose.yml
```diff
@@ -51,16 +124,20 @@ services:
...
-        SERVERS_RED_LOG_FILE_PATH: "/logfiles/${SERVERS_RED_LOG_FILE}"
+        SERVERS_BLUE_LOG_FILE_PATH: "${SERVERS_BLUE_LOG_FILE_PATH:-/logfiles/${SERVERS_BLUE_LOG_FILE}}"
+        SERVERS_GREEN_LOG_FILE_PATH: "${SERVERS_GREEN_LOG_FILE_PATH:-/logfiles/${SERVERS_GREEN_LOG_FILE}}"
+        SERVERS_RED_LOG_FILE_PATH: "${SERVERS_RED_LOG_FILE_PATH:-/logfiles/${SERVERS_RED_LOG_FILE}}"
         REDIS_SOCKET_DIR: "${REDIS_SOCKET_DIR}"
         FAKE_LOGS: "${FAKE_LOGS:-false}"
         DEBUG_MODE: "${DEBUG_MODE:-false}"
       volumes:
         - type: bind
           source: "${LOG_SOURCE_PATH}"
```
jamesjamail: This bind is unconditional, so the headless path still needs a real EverQuest log directory on the host.

The README says the pinned image means "users do not need an EverQuest installation or separate asset configuration" — but a user without one copies `.env.example`, whose `LOG_SOURCE_PATH` is the placeholder `C:\\Program Files (x86)\\Sony\\EverQuest\\Logs\\`, and Compose refuses to create a bind mount for a source that doesn't exist. The stack never starts, even though every `SERVERS_*_LOG_FILE_PATH` now points at `/p99-logger`.

Could the `/logfiles` bind move into a profile of its own, or default to something innocuous (`LOG_SOURCE_PATH:-./ci-logs`) so the collector path stands alone?

src/lib/parser/monitorLogs.ts
```diff
@@ -48,17 +48,90 @@ export function generateAuctionKey(auctionText: string) {
...
+	if (auctionMatch) {
+		return {
+			kind: 'auction',
+			playerName: auctionMatch[1],
+			text: auctionMatch[2],
+		};
+	}
+
 	const linkMatch = data.match(
 		/(\w+) says? out of character, 'Link me: ([0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12})'/,
```
jamesjamail: Minor: this inline pattern and the new `LINK_CODE` constant above carry the same 60-character UUID body, so the link-code format is now specified twice in one function.

If the accepted shape ever changes, updating one and missing the other means JSONL records and identical text-log lines silently disagree about what counts as a link request. Interpolating `LINK_CODE`'s source into this regex, or extracting the UUID part as a shared fragment, would keep them in step.

jamesjamail (commented):
Nice piece of work — this removes the EverQuest-client requirement entirely, which is the last hard prerequisite for running the bot.

The parser change reads well. `parseLogLine` factors both formats behind one discriminated union, the text-log path keeps its exact prior behaviour, and the JSONL branch validates `type`, `sender` and `text` before trusting a record. The container hardening is thorough too — `cap_drop: ALL`, read-only rootfs, `no-new-privileges`, a non-root `user:`, and `create_host_path: false` so a missing credential file fails loudly instead of being created as a directory. Credentials are kept out of both Git and the Docker context.

Three things below. The two Compose ones are the same underlying theme: the collector path is wired as an *addition* to the existing bind-mount setup rather than an alternative to it, so the "no EverQuest installation" path doesn't quite hold end to end yet.

Note this targets `dx-improvements-substantive` rather than `main`. #107 has since had a CI failure fixed (a test of mine that depended on build output the `test` job runs before producing), so this will want a re-run against the updated base.

docker-compose.yml
rm-you: Fixed in `6314c22`. The initializer now runs without a profile, and the bot depends on its successful completion; only the collectors remain profile-gated. I tested both configurations with a deliberate three-second initializer delay and verified all three log files exist with the expected permissions before the bot starts.

docker-compose.yml
rm-you: Changed both `.env.example` and the Compose fallback to `./logs` in `6314c22`, and excluded that directory from Git and the Docker build context. Headless setups can keep that default; existing client-log deployments can still supply their own path. Verified startup without an existing log directory and checked the explicit legacy-path override.

src/lib/parser/monitorLogs.ts
rm-you: Done in `6314c22`: both expressions now use a shared `UUID_PATTERN`, preserving their existing anchors and capture groups. The native text and JSONL parser tests pass with the rest of the 619-test suite. GitHub CI is also green.

PR #122, opened 2026-09-07
Harden long-running collector operations

rm-you:
## Summary

- Cap Docker stdout/stderr with the rotating `local` driver.
- Rotate Green/Blue JSONL files with a non-root logrotate sidecar.
- Require positive rotation sizes with an explicit unit and positive check intervals, including validation of zero-padded values.
- Report failed rotation passes and retry after the normal interval while preserving rotation state.
- Treat historical pricing network/JSON failures as optional enrichment, keeping auction streams running without Discord error reports when PigParse is unavailable.

## Defaults

- Docker logs: 5 files × 10 MB per container.
- JSONL: rotate at 100 MB, retain 5 compressed archives, check every 5 minutes.

## Validation

- Formatting, lint, type checks, coverage thresholds, and build passed on Node 24/Linux.
- Full unit suite: 641 passed, including 21 process-level retention tests.
- The new tests reproduced the validation and retry defects before the fix.
- Updated retention image built successfully; an isolated non-root container with synthetic logs verified actual copytruncate/compression, accepted size units, rejection of invalid settings, and recovery with preserved state after a simulated rotation failure.

PR #123 is stacked on this branch.

### DIFF
--- a/p99-logger/retention-entrypoint.sh
+++ b/p99-logger/retention-entrypoint.sh
@@ -1,0 +1,38 @@
+#!/bin/sh
+set -eu
+
+max_size="${P99_JSONL_MAX_SIZE:-100M}"
+rotate_count="${P99_JSONL_ROTATE_COUNT:-5}"
+interval="${P99_JSONL_ROTATE_INTERVAL_SECONDS:-300}"
+
+case "$max_size" in
+	*[kKmMgG]) size_number="${max_size%?}" ;;
+	*) size_number="$max_size" ;;
+esac
+case "$size_number" in
+	''|*[!0-9]*) echo "P99_JSONL_MAX_SIZE must be a size such as 100M" >&2; exit 1 ;;
+esac
+case "$rotate_count" in
+	''|*[!0-9]*) echo "P99_JSONL_ROTATE_COUNT must be a non-negative integer" >&2; exit 1 ;;
+esac
+case "$interval" in
+	''|*[!0-9]*|0) echo "P99_JSONL_ROTATE_INTERVAL_SECONDS must be a positive integer" >&2; exit 1 ;;
+esac
+
+cat > /tmp/logrotate.conf <<EOF
+/data/green/chat.jsonl /data/blue/chat.jsonl {
+	size $max_size
+	rotate $rotate_count
+	missingok
+	notifempty
+	copytruncate
+	compress
+}
+EOF
+
+echo "JSONL retention active: check every ${interval}s, rotate at ${max_size}, keep ${rotate_count} compressed archives"
+
+while true; do
+	logrotate --state /tmp/logrotate.status /tmp/logrotate.conf
+	sleep "$interval"
+done
--- a/p99-logger/retention-entrypoint.sh
+++ b/p99-logger/retention-entrypoint.sh
@@ -5,19 +5,30 @@ max_size="${P99_JSONL_MAX_SIZE:-100M}"
 rotate_count="${P99_JSONL_ROTATE_COUNT:-5}"
 interval="${P99_JSONL_ROTATE_INTERVAL_SECONDS:-300}"
 
+# Check decimal digits without shell arithmetic or its leading-zero semantics.
+is_positive_integer() {
+	case "$1" in
+		''|*[!0-9]*) return 1 ;;
+		*[1-9]*) return 0 ;;
+		*) return 1 ;;
+	esac
+}
+
 case "$max_size" in
-	*[kKmMgG]) size_number="${max_size%?}" ;;
-	*) size_number="$max_size" ;;
+	*[kKMG]) size_number="${max_size%?}" ;;
+	*) size_number= ;;
 esac
-case "$size_number" in
-	''|*[!0-9]*) echo "P99_JSONL_MAX_SIZE must be a size such as 100M" >&2; exit 1 ;;
-esac
+if ! is_positive_integer "$size_number"; then
+	echo "P99_JSONL_MAX_SIZE must be a positive integer followed by k, K, M, or G, such as 100M" >&2
+	exit 1
+fi
 case "$rotate_count" in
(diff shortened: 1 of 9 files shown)


p99-logger/retention-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+	*) size_number="$max_size" ;;
+esac
+case "$size_number" in
+	''|*[!0-9]*) echo "P99_JSONL_MAX_SIZE must be a size such as 100M" >&2; exit 1 ;;
+esac
+case "$rotate_count" in
+	''|*[!0-9]*) echo "P99_JSONL_ROTATE_COUNT must be a non-negative integer" >&2; exit 1 ;;
+esac
+case "$interval" in
+	''|*[!0-9]*|0) echo "P99_JSONL_ROTATE_INTERVAL_SECONDS must be a positive integer" >&2; exit 1 ;;
```
jamesjamail: This rejects the literal `0` but not `00`, and `sleep 00` returns immediately — I measured 0.00s real.

So `P99_JSONL_ROTATE_INTERVAL_SECONDS=00` passes validation and turns the loop below into a busy loop: `logrotate` running continuously, pinning a core and hammering the shared volume, with `restart: unless-stopped` keeping it there. Zero-padding a number in a `.env` isn't a stretch.

Stripping leading zeros before the check, or comparing numerically (`[ "$interval" -gt 0 ]` once you know it's all digits), covers it.

p99-logger/retention-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+set -eu
+
+max_size="${P99_JSONL_MAX_SIZE:-100M}"
+rotate_count="${P99_JSONL_ROTATE_COUNT:-5}"
+interval="${P99_JSONL_ROTATE_INTERVAL_SECONDS:-300}"
+
+case "$max_size" in
+	*[kKmMgG]) size_number="${max_size%?}" ;;
+	*) size_number="$max_size" ;;
```
jamesjamail: The `*)` branch lets a unit-less number through, and logrotate reads a bare `size` value as **bytes**.

`P99_JSONL_MAX_SIZE=100` — a natural thing to write when the documented value is `100M` — therefore passes validation and configures rotation at 100 bytes. Every five-minute pass then rotates and compresses, churning the five-archive window continuously and truncating the file the bot is tailing. `0` gets through the same way.

The error message on the next line already says "a size such as 100M", so requiring the suffix would just make the code agree with it: drop the `*)` fallback and let anything without `[kKmMgG]` fail.

p99-logger/retention-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+	notifempty
+	copytruncate
+	compress
+}
+EOF
+
+echo "JSONL retention active: check every ${interval}s, rotate at ${max_size}, keep ${rotate_count} compressed archives"
+
+while true; do
+	logrotate --state /tmp/logrotate.status /tmp/logrotate.conf
```
jamesjamail: Minor: `set -e` is still in effect here, so a single non-zero exit from `logrotate` ends the loop and the container.

`restart: unless-stopped` brings it back, so this is self-healing rather than fatal — but the state file lives in the tmpfs `/tmp`, so each restart also discards `logrotate.status` and resets the bookkeeping. A transient failure (a compress interrupted, one path briefly unreadable while the collector recreates it) costs more than it needs to.

`logrotate ... || echo 'rotation pass failed; retrying next interval' >&2` would keep the loop alive and make the failure visible in the container log you're now capping.

jamesjamail (commented):
This is a good shape overall — capping Docker's stdout with an anchored `local` driver, and demoting PigParse to optional enrichment so an outage stops spamming the error channel rather than interrupting auction streams, are both clearly right.

The sidecar is properly locked down too: `read_only`, `cap_drop: ALL`, `no-new-privileges`, non-root, size-capped tmpfs, and running as the same uid that owns the files so `copytruncate` can actually write them.

I checked the `copytruncate` claim in the README and it holds — `tail` resets `currentCursorPos` when the file shrinks (`node_modules/tail/lib/tail.js:245`), so the bot keeps following across a rotation without a restart. Documenting the copy-window line loss rather than glossing it is the right call.

Three things in `retention-entrypoint.sh` below. **None of them affect the shipped defaults** — `100M` / `5` / `300` all behave correctly — so none are merge blockers. They're all misconfiguration or transient-failure paths, and they're cheap to fix while the file is new. I ran the validation logic directly rather than reading it, so the cases below are confirmed rather than suspected.

p99-logger/retention-entrypoint.sh
rm-you: Fixed in aea5f62. Validation now requires decimal digits with at least one nonzero digit, so `0`, `00`, and `000` are all rejected while positive zero-padded values remain valid. Added process-level regression tests that verify rejection occurs before any rotation files are created.

p99-logger/retention-entrypoint.sh
rm-you: Fixed in aea5f62. Sizes now require a positive whole number and a logrotate-supported suffix (`k`, `K`, `M`, or `G`). This rejects bare `100`, zero, and suffixed zero values such as `0M` and `00M`. The shipped logrotate also rejects lowercase `m` and `g`, so validation now catches those at startup. Updated the examples/docs and verified accepted sizes with real logrotate in an isolated non-root container.

p99-logger/retention-entrypoint.sh
rm-you: Fixed in aea5f62. A failed rotation now reports `rotation pass failed; retrying next interval` and proceeds to the normal sleep instead of exiting. The regression test verifies failure reporting, the delay before retry, and preservation of the state file. An isolated image check also confirmed a successful second pass after a simulated failure.

PR #123, opened 2026-09-07
Fix development configuration and publish validated images

### LINKED ISSUE #109: Preserve custom socket database URLs in the development launcher
Nonblocking follow-up from #107, reproduced at `855085a`.

`scripts/dev.mjs` treats any `.env` DATABASE_URL containing `?host=` or `&host=` as container-only and replaces it with the default localhost TCP URL. Unix sockets are also valid for host development, so this discards genuine overrides.

For example, `DATABASE_URL=postgresql://reviewer:example@localhost/custom_db?host=/tmp/local-postgres` is replaced with the default `tunnelquestbot` database. The new migration command receives the replacement too and can therefore target a different database from the one selected in configuration.

Distinguish the shipped container sample from explicit user configuration, or provide an explicit development database URL setting. Add a regression case for a custom socket URL in `.env`.

The documented setup works. Exporting DATABASE_URL in the shell currently preserves a custom override. Validation used the real launcher with subprocesses intercepted and a pg client connected to an isolated Unix-socket listener; no real database migrations were run.

### LINKED ISSUE #110: Expand fake-log configuration before choosing development processes
Nonblocking follow-up from #107, reproduced at `855085a`.

The launcher reads raw dotenv values while the bot expands them. This configuration still makes the launcher skip logFaker even though the bot enables fake logs:

```dotenv
USE_FAKE_LOGS=true
FAKE_LOGS=${USE_FAKE_LOGS}
```

Resolve the effective environment using the same expansion and precedence as the children before deciding which processes to start. Add a regression case that checks the launcher and bot agree for an interpolated FAKE_LOGS value.

Literal true/false values work, including the shipped example. Validation ran the unchanged launcher with subprocess creation intercepted, then loaded the same configuration using the bot's actual dotenv expansion and configuration parser.

### LINKED ISSUE #120: Exercise logFaker in the smoke job
Deferred from #107, where @rm-you noted the smoke job's claim is broader than what it actually runs.

`docker-entrypoint.sh` short-circuits on `SMOKE_TEST` before reaching the `FAKE_LOGS` branch, so the smoke fixture sets `FAKE_LOGS=true` and nothing ever consumes it. `logFaker.js` is never exercised, despite being on the path a contributor uses for all host development.

Closing this means reordering that branch so the fake-log setup runs before the smoke short-circuit. The risk is that the `startup migration invariants` tests in `test/config.test.ts` pin the current ordering — specifically that `apply_migrations` precedes both `$FAKE_LOGS` and `$DEBUG_MODE`, and that the smoke branch precedes `$FAKE_LOGS`. Any reordering has to keep migrations first and update those assertions deliberately rather than incidentally.

### LINKED ISSUE #121: Publish the exact image the smoke job tested
Deferred from #107, raised by @rm-you.

The image is built twice per push to `main`: `docker compose build tunnelquestbot` in the `smoke` job, with no buildx and no GHA cache, and then again in `build-image-and-push` via `docker/build-push-action`. They are separate builds, so the artifact that gets published is not the one that passed smoke.

The fix is to build and push by digest first, run smoke against that digest, then tag it — which removes the duplicate build as well as the gap.

Worth doing after #108, which adds `p99-logger-init` and two collector services to the compose stack and changes what the smoke job brings up.

Related: `build-image-and-push` currently tags `:latest` and `:${{ github.sha }}`. The sha tag was added in #107 to give a rollback target; a digest-based flow should keep both.

rm-you:
Stacked on #122 (`codex/p99-operational-hardening`).

The development launcher currently replaces custom Unix-socket database URLs and chooses whether to start logFaker before expanding dotenv references. It now resolves configuration once for migrations and child processes, preserves custom URLs/socket directories, and translates only the shipped Compose template to host TCP while retaining the configured database and credentials.

Container startup now runs the real fake-log writer once after migrations and before the smoke check. Missing or failing writer code stops startup; smoke validation checks nonempty Blue, Green, and Red log files. Its disposable Compose override uses independent container names, temporary configuration, and no build instruction, so it can coexist with a running deployment.

Both PR and main image jobs use one reusable build/smoke workflow. PR checks load a local image with read-only repository permissions. After unit and integration checks pass on main, the workflow pushes an untagged candidate digest, pulls that exact digest for smoke testing, and assigns the commit-SHA and latest tags only after smoke succeeds. Publishing never rebuilds the image.

Validation:
- formatting, lint, type checks, compilation, and unit tests with coverage thresholds passed on Linux
- all 55 integration tests passed against disposable Postgres/Redis containers
- Actionlint and POSIX shell syntax checks passed
- production-image Compose smoke passed; an injected writer failure was rejected before doctor startup
- local registry round trip verified that both promoted tags retain the exact digest that passed smoke, including its provenance manifest
- reviewed staged files: fixtures contain synthetic credentials and no collected game logs

Fixes #109, fixes #110, fixes #120, fixes #121.


docker-entrypoint.sh
```diff
@@ -68,6 +68,12 @@ apply_migrations() {
 
 apply_migrations
 
+# Create and write the real fake-log files synchronously. Both smoke mode and
+# normal startup fail if the writer or its assets are missing or unwritable.
+case "$FAKE_LOGS" in
+	[tT]*) node ./build/lib/parser/logFaker.js --once ;;
```
jamesjamail: Non-blocking, but this runs on normal startup too, not just smoke — and both consequences look unintended.

With `FAKE_LOGS=true` and `SMOKE_TEST` unset, control reaches line 93, which starts `logFaker.js &`. That invocation calls `appendFakeLogs()` immediately before setting its interval, so the generator writes an initial batch twice within the same second.

More relevant: `set -e` on line 2 means a non-zero exit here aborts the entrypoint before the bot starts, where the backgrounded call on line 93 previously let startup continue. Under `restart: always`, an unwritable `build/lib/fakeLogs` becomes a crash loop rather than a degraded run. The comment argues for failing when the writer or its assets are missing, which is right for smoke — I'm less sure it's right for a running deployment, where losing generated logs is better than losing the bot.

Gating this on `SMOKE_TEST` would keep the guarantee where it's needed without changing production startup behaviour.

jamesjamail (commented):
This is the strongest of the three. Publishing by digest — push untagged, pull that exact digest, smoke it, then tag `latest` and the SHA only on success — genuinely closes #121 rather than narrowing it, and folding both paths into one reusable workflow means the PR path and the publish path can't drift.

`ci-smoke.sh` is well built: temp dir, `!reset null` so it never touches a developer's `.env` or container names, `pull_policy: never` so it can't silently validate a registry image instead of the local one, and cleanup on EXIT/INT/TERM.

I checked the `resolveDevEnv` rewrite against `dotenv-expand@13` rather than assuming, since it replaces code I wrote. Shell precedence survives: `expand()` short-circuits on `if (processEnv[key] && processEnv[key] !== value)`, and the write-back at the end preserves the exported value. Empty-value and missing-key cases behave as before. Using real expansion instead of my defaults-only `applied` is a straight improvement — migrations, the bot and logFaker now see identical resolved values, which is what my version couldn't guarantee.

One non-blocking note inline, plus one thing worth stating explicitly rather than leaving implicit:

The container-URL translation narrowed from a `[?&]host=` test to an exact template match plus `DB_SOCKET_DIR === '/dbsocket'`. Anyone who changes `DB_SOCKET_DIR` now gets the socket URL on the host — the original failure, behind a narrower door. `config.ts` rejects the unresolved URL with a clear message rather than failing obscurely, and you documented it in both `.env.example` and the README, so I read it as a deliberate trade rather than an oversight. Flagging only so the decision is on the record.

docker-entrypoint.sh
rm-you: Fixed in 114ac83 and a5eeecb. The synchronous `logFaker.js --once` invocation now runs only inside smoke mode. Normal startup launches the background generator once and continues to the bot even if that generator fails.

Added process-level tests for normal startup with both a successful and failing writer, retained fatal writer failures in smoke mode, and updated the existing startup-order assertion. Both startup test suites passed (38 tests); the updated entrypoint also passed the built-image smoke checks on #125.

jamesjamail:
<3

PR #125, opened 2026-09-07
Use SQLite storage and migrate existing PostgreSQL data

rm-you:
## Summary

TQB uses ordinary Prisma CRUD and relations, so its current production queries do not require PostgreSQL. Switch the default stack to a persistent SQLite file and remove the PostgreSQL service from normal operation. Redis remains unchanged.

- Use Prisma's SQLite adapter with foreign keys, WAL journaling and a five-second busy timeout. Preserve keys, indexes, cascades, nullable unique keys, enum/length checks and UUID lookup behavior.
- Keep host development and container migrations on the same database path and timestamp representation. Replace PostgreSQL-specific foreign-key error matching with an idempotent user upsert.
- Share one native SQLite library between Prisma and the importer, with a regression test preventing duplicate driver installations.

Migration failures report the stage, table and a safe cause while omitting connection URLs and database values. Diagnostic tests include a real rejected-password failure and the production CLI.

## Existing installations

Follow [the migration guide](https://github.com/jamesjamail/TunnelQuestBot/blob/codex/sqlite-storage/docs/sqlite-migration.md). Stop the old bot and take a backup, then start with `docker-compose.postgres-migration.yml`. Startup automatically copies the original PostgreSQL database into SQLite before starting the bot.

The importer copies all five tables in bounded batches, preserves IDs and sequence high-water marks, and verifies counts, foreign keys and integrity before committing. Failed imports roll back; successful imports are recorded and skipped on restart. The source database remains intact. Old PostgreSQL settings without the migration override prevent startup instead of silently using an empty SQLite database.

SQLite is intended for one bot instance with local persistent storage. It trades PostgreSQL's concurrent-writer and multi-host capabilities for a simpler deployment. The guide covers backup, rollback and future migration constraints.

## Validation

- Formatting, lint, type checking and build.
- 693 unit tests with coverage thresholds; 59 integration tests.
- Real disposable PostgreSQL-to-SQLite import, including rollback/retry, all five tables, multiple batches, timestamps, nullable fields, IDs and restart without a reachable source.

- Both production-image Compose checks passed: fresh startup and the original PostgreSQL socket/volume cutover, including a restart with PostgreSQL stopped. CI now runs both before promoting an image.

Stacked on #123; merge that PR first.


src/prisma/import-postgres.ts
```diff
@@ -0,0 +1,189 @@
...
+			process.env.DATABASE_URL ?? 'file:./data/tunnelquestbot.db',
+		)
+			.then((counts) =>
+				console.log(
+					counts
+						? `[database] PostgreSQL import complete: ${JSON.stringify(counts)}`
+						: '[database] PostgreSQL import already completed; using SQLite',
+				),
+			)
+			.catch(() => {
```
jamesjamail: This discards the error it is reporting on, which matters more here than almost anywhere else in the codebase — it is the failure path of a one-shot migration of someone's production data.

Every distinct cause currently collapses to the same three lines: a wrong host or rejected credentials in `POSTGRES_MIGRATION_URL`, a `query_timeout` on a large table, `Refusing to import into a nonempty SQLite database`, `Row-count verification failed for Watch`, `Imported foreign keys are inconsistent`, `Imported SQLite database failed its integrity check`, and `DATABASE_URL must be a local SQLite file URL` thrown out of `prepareSqliteUrl`. Each of those already carries the distinguishing text, and the handler drops it before printing.

The operator is then left with a bot that will not start — `set -e` makes the non-zero exit fatal at entrypoint line 69 — and a suggestion to check three things, one of which is usually irrelevant. Telling a typo in a connection string from a failed integrity check is the whole question at that moment.

`.catch((error) => { console.error(...); console.error(error); ... })` would keep the guidance and add the cause. This is the same thing you flagged in my `setup.mjs` on #107, where only `error.status` survived and a DNS failure rendered as advice about guild ids — the guidance is good, it just needs the error alongside it.

jamesjamail (commented):
Reviewed this one more carefully than the last two, since it moves live data.

The importer is the part I spent the most time on and it holds up. `BEGIN IMMEDIATE` before touching the source, refusing a nonempty target, `REPEATABLE READ READ ONLY` on the source, per-table count verification on both sides, `foreign_key_check` and `quick_check` before commit, the `DataMigration` row written inside the same transaction so a re-run is a no-op, and `ROLLBACK` on the source so it is never altered. Table order is FK-safe, and carrying sequence high-water marks — including IDs of deleted rows — is a detail that is easy to miss and expensive later.

Things I checked specifically rather than assumed:

- **Persistence.** `sqlite-data:/data` with `DATABASE_URL` defaulting to `file:/data/tunnelquestbot.db`, so an image update does not take the database with it.
- **The native module.** `better-sqlite3` has to compile against musl; the build stage adds `python3 make g++` and rebuilds, and both stages share the same Alpine base, so the binary matches.
- **Query semantics.** The two SQLite traps I expected both miss us. Nothing uses Prisma's `mode: 'insensitive'` (#124 filters in JS), and there are no `contains`/`startsWith`/`endsWith` filters, so the `LIKE` case-sensitivity difference between the engines cannot change a result. No raw SQL outside test setup.
- **Constraint fidelity.** `VarChar(255)` → `length()` CHECKs, enums → `IN` CHECKs, `@db.Uuid` → a real structural CHECK plus `COLLATE NOCASE`, and every index, unique key and cascade preserved.
- **Host development is fully converted** — `dev:deps` starts only Redis, the `postgres` service is gone, and the compose-template translation from #123 was removed rather than left behind as dead code.

The docs are honest in the places that matter: the WAL caveat on naive file copies, and rollback losing post-cutover changes.

One finding inline. Two notes that need no action, just visibility:

**Enum changes get expensive.** A CHECK constraint cannot be altered, so adding a fourth server becomes a rebuild migration across `Watch`, `BlockedPlayer` and `PlayerLink` — where on PostgreSQL it was `ALTER TYPE ... ADD VALUE`. `config.ts` derives `SERVER_NAMES` and its env keys from that enum, so the code side stays a one-liner while the schema side no longer is.

**`attemptAndCreateUserIfNeeded` now upserts unconditionally.** Correct given SQLite's foreign-key errors do not name the column. I checked both call sites — `createWatch` and `insertPlayerLink` — so this does not add a write to any read-only command on a single-writer database.

src/prisma/import-postgres.ts
rm-you: Fixed in 7343a3c. Migration failures now identify the operation and, during a table copy, the table, followed by a safe cause. For example: `connecting to PostgreSQL: PostgreSQL authentication failed (28P01)`. Application-owned validation messages retain their specific nonempty-target, row-count, foreign-key and integrity-check reasons.

Driver errors are classified using known codes and exact timeout messages. Raw driver messages, URLs, SQL, row values and stacks are omitted; an unfamiliar error retains the stage/table with an explicit unclassified-failure category. This keeps the useful cause without risking credentials or production data in container logs.

Added diagnostic/privacy tests, a real PostgreSQL rejected-password test, and a built-image CLI check that verifies useful output without either connection URL appearing. Validation passed: 693 unit tests with coverage, 59 integration tests, and both container smoke checks, including import/restart. Also rebased this PR onto the updated #123.

jamesjamail:
<3

PR #126, opened 2026-09-08
Add guarded SQLite cutover tooling

rm-you:
## Summary
- update the production-only Compose bundle for SQLite and Redis health checks
- modernize Windows deployment scripts for Docker Compose v2 and visible failures
- block routine startup until the one-time PostgreSQL cutover completes
- add a PowerShell 5.1-compatible cutover with private SQL/.env backups, independent row reconciliation, integrity checks, dependency startup, and failure recovery
- document first-time cutover, rollback, and routine operations

## Validation
- parsed and exercised environment rewrite logic with Windows PowerShell 5.1
- validated regular and migration Compose configurations
- asserted persistent SQLite, Redis health dependency, and migration overlay invariants
- completed an independent deployment safety review and addressed all findings

## Deployment dependency
Merge and promote the SQLite-capable production image from #125 before merging or using this deployment update.

### DIFF
--- a/assert-sqlite-ready.ps1
+++ b/assert-sqlite-ready.ps1
@@ -9,17 +9,17 @@ try {
         throw 'Missing .env in the deployment directory'
     }
     $envText = [IO.File]::ReadAllText((Resolve-Path '.env'))
-    $matches = [regex]::Matches(
+    $databaseMatches = [regex]::Matches(
         $envText,
         '(?m)^(?!\s*#)\s*DATABASE_URL\s*=\s*(.+?)\s*$'
     )
-    if ($matches.Count -ne 1) {
-        throw "Expected exactly one active DATABASE_URL in .env; found $($matches.Count)"
+    if ($databaseMatches.Count -ne 1) {
+        throw "Expected exactly one active DATABASE_URL in .env; found $($databaseMatches.Count)"
     }
-    if ($matches[0].Groups[1].Value -match '^\s*[''"]?postgres(?:ql)?:') {
+    if ($databaseMatches[0].Groups[1].Value -match '^\s*[''"]?postgres(?:ql)?:') {
         throw 'PostgreSQL cutover is pending. Run migrate-postgres-to-sqlite.bat instead.'
     }
-    if ($matches[0].Groups[1].Value -notmatch '^\s*[''"]?file:') {
+    if ($databaseMatches[0].Groups[1].Value -notmatch '^\s*[''"]?file:') {
         throw 'DATABASE_URL must be a persistent SQLite file URL'
     }
 } catch {
--- a/assert-sqlite-ready.ps1
+++ b/assert-sqlite-ready.ps1
@@ -1,0 +1,28 @@
+#requires -Version 5.1
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+Set-Location $PSScriptRoot
+
+try {
+    if (-not (Test-Path '.env')) {
+        throw 'Missing .env in the deployment directory'
+    }
+    $envText = [IO.File]::ReadAllText((Resolve-Path '.env'))
+    $matches = [regex]::Matches(
+        $envText,
+        '(?m)^(?!\s*#)\s*DATABASE_URL\s*=\s*(.+?)\s*$'
+    )
+    if ($matches.Count -ne 1) {
+        throw "Expected exactly one active DATABASE_URL in .env; found $($matches.Count)"
+    }
+    if ($matches[0].Groups[1].Value -match '^\s*[''"]?postgres(?:ql)?:') {
+        throw 'PostgreSQL cutover is pending. Run migrate-postgres-to-sqlite.bat instead.'
+    }
+    if ($matches[0].Groups[1].Value -notmatch '^\s*[''"]?file:') {
+        throw 'DATABASE_URL must be a persistent SQLite file URL'
+    }
+} catch {
+    Write-Host "ERROR: $($_.Exception.Message)" -ForegroundColor Red
+    exit 1
+}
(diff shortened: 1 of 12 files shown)


migrate-postgres-to-sqlite.ps1
```diff
@@ -0,0 +1,319 @@
...
+    $botState = docker inspect tunnelquestbot --format '{{.State.Status}}'
+    Assert-LastExitCode 'Bot status check'
+    $restartCount = [int](docker inspect tunnelquestbot --format '{{.RestartCount}}')
+    Assert-LastExitCode 'Bot restart check'
+    if ($botState.Trim() -ne 'running' -or $restartCount -ne 0) {
+        docker logs --tail 100 tunnelquestbot
+        throw "The SQLite bot is not stable (status=$botState, restarts=$restartCount)"
+    }
+    $sqliteState = Test-SqliteImport
+    Compare-RowCounts $postgresCounts $sqliteState.counts
```
jamesjamail: This compares a snapshot taken before the cutover against counts read while the bot is live, and the failure lands after the point of no return.

Line 258 restarts the bot on SQLite, so it connects to Discord and starts serving commands. Ten seconds later line 270 re-reads the counts and this line compares them to `$postgresCounts` from line 227. A single `/watch`, `/unwatch` or player link in that window legitimately adds a row — each inserts through `attemptAndCreateUserIfNeeded` — and `Compare-RowCounts` throws `Row-count mismatch for Watch`.

The catch block then stops the bot and rewrites `.env` back to PostgreSQL. But the import has already committed, including the `postgres-to-sqlite-v1` row in `DataMigration`. Re-running takes the `already completed` path, re-reads the now-larger counts, and fails the same comparison again. The operator is stuck mid-cutover on a live deployment, with the script's own rollback having moved `.env` backwards while the data moved forwards.

The comparison at line 233 has the same shape, since the bot is also running under the migration Compose config.

Two ways out, either fine: do the strict equality check before the bot serves traffic and make the post-restart check a liveness check only; or, once the marker is present, treat SQLite >= PostgreSQL as success and only fail on a shortfall. The second keeps the post-restart verification meaningful while making it immune to normal traffic.

assert-sqlite-ready.ps1
```diff
@@ -0,0 +1,28 @@
...
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+Set-Location $PSScriptRoot
+
+try {
+    if (-not (Test-Path '.env')) {
+        throw 'Missing .env in the deployment directory'
+    }
+    $envText = [IO.File]::ReadAllText((Resolve-Path '.env'))
+    $matches = [regex]::Matches(
```
jamesjamail: Latent rather than live, but worth renaming: `$matches` shadows PowerShell's automatic `$Matches`, which line 19's `-match` rewrites with a capture-group hashtable whenever it succeeds.

Today the script is fine only because line 20 throws immediately, so line 22 never reads the clobbered value on that path. Any later edit that inserts a check between them, or softens the throw into a warning, leaves line 22 evaluating `$matches[0]` as a plain string — and with `Set-StrictMode -Version Latest`, the missing `.Groups` property becomes a hard failure in the one script that gates every production start.

`migrate-postgres-to-sqlite.ps1` already avoids this by naming its collection `$databaseMatches`; the same name here would make the two consistent.

I could not run PowerShell to reproduce this, so please sanity-check the failed-match behaviour rather than taking it on faith.

jamesjamail (commented):
Reviewed this one closely — it is the piece that touches live production data.

The safety engineering here is better than most migration tooling I have seen. Validating the dump by size *and* the `-- PostgreSQL database dump complete` marker rather than just checking the file exists; taking row counts independently through `psql` so the importer's own arithmetic is never the thing verifying the importer; checking `foreign_key_check`, `quick_check` and the migration marker on the destination; requiring a typed `MIGRATE`; retaining the PostgreSQL container and volume; and a catch block that restores `.env`, restarts PostgreSQL and leaves the bot stopped rather than half-migrated. `docker exec postgres sh -ec 'umask 077; pg_dump ...'` is a nice detail — the dump is never briefly world-readable.

I also confirmed the production Compose keeps `sqlite-data:/data` with `DATABASE_URL` defaulting into it, so an image update cannot take the database with it. And gating `start.bat` behind `assert-sqlite-ready.ps1` means a routine start refuses to run while a PostgreSQL URL is still configured, rather than silently booting onto an empty SQLite file.

Two findings inline. The first is the one I would want fixed before running this against the real deployment; the second is latent rather than live.

One note needing no action: I could not run PowerShell here, so the second finding is reasoned from the language semantics rather than reproduced. Worth a quick check on the NUC rather than taking my word for it.

migrate-postgres-to-sqlite.ps1
rm-you: Fixed in 6e6a46c using the offline verification approach. Migrations and the importer now run in a one-shot `docker compose run --no-deps` container while the bot stays stopped. Independent row-count equality, foreign-key, integrity and import-marker checks complete before the bot starts. After startup, we check stability and integrity without comparing live counts to the PostgreSQL snapshot; normal additions and deletions are both allowed.

Recovery also respects that boundary: once SQLite startup has been attempted, a failure leaves the bot stopped with SQLite still configured, since it may have accepted writes. The README explains repair and rollback at that point.

Validation passed: 13 Windows PowerShell 5.1 scenarios, including live additions/deletions and failures before/after activation. A separate disposable-container test imported and verified all five tables, including 512 watches, with the bot service never started and PostgreSQL unchanged.

assert-sqlite-ready.ps1
rm-you: Fixed in 6e6a46c: renamed the collection to `$databaseMatches`, matching the migration script.

I checked the behavior in Windows PowerShell 5.1: a failed `-match` leaves the MatchCollection unchanged; a successful match replaces it with a Hashtable. That confirms the existing throw protected today's successful-match path. The regression suite also exercises the startup guard with SQLite, PostgreSQL, invalid and duplicate database settings.

jamesjamail:
<3

PR #132, opened 2026-09-16
minimal marketplace via `watch matching`

### LINKED ISSUE #47: Prompt users when adding a watch to be connected with other TQB users who are buying/selling same item
-alter users table to add AGREED boolean field
-alter watch table to add opt-in boolean field

-when a user adds a watch, they are prompted:

"would you like to be connected with other TQB users buying/selling (depending on watch type) of <ITEM>?  Opting in to this feature will also allow other TQB users to contact you via Discord"

Potentially add commands so users can check who's buying/selling what.

ChuckHend:
## What changed

Create a basic marketplace. I /WTB an item and you /WTS an item, connect the two of us so we can trade.

Closes https://github.com/jamesjamail/TunnelQuestBot/issues/47

## Checks

- [x] `npm run check` passes (lint + typecheck + unit tests)
- [x] `npm run test:integration` passes, or this change doesn't touch `src/prisma/`
- [x] New behaviour has a test; changed parser behaviour has a case in
      `parser.test.ts` or `unknownItems.test.ts`


src/prisma/dbExecutors/marketplace.ts
```diff
@@ -0,0 +1,304 @@
...
+		if (isPrismaUniqueConstraintViolation(error)) {
+			return null;
+		}
+		throw error;
+	}
+}
+
+type WatchPair = { wtbWatch: WatchWithUser; wtsWatch: WatchWithUser };
+
+function buildCandidatePairs(
```
jamesjamail, line 97: **Block lists are ignored when pairing traders.**

`findEligibleWatchesByType` filters on watchType/active/`isPublicallyTradeable`/snooze, and `buildCandidatePairs` excludes only self-pairs and price-incompatible pairs. Neither consults `BlockedPlayer` or `BlockedPlayerByWatch`.

So: user runs `/block NAUGHTYSELLER GREEN` after a bad trade. That account has an opted-in watch for the same item on Green. The pair is recorded and both users are DMed each other's handle. `shouldUserBeNotified` enforces both block types on the auction path — the feature that most directly puts users in contact is the one ignoring the mechanism for saying "don't contact me."

The namespace mismatch is why this isn't a one-line fix: blocks key on in-game character name, matching keys on `discordUserId`, bridged only by `PlayerLink`. Given how few users have linked, a `discordUserId`-keyed marketplace block kept separate from character blocks is the way to go — and it has to be **mutual** (drop the pair if either side blocked the other), since a one-directional block would leave the blocked party still seeing you and still able to DM you directly.

src/prisma/dbExecutors/watch.ts
```diff
@@ -58,6 +68,9 @@ export async function upsertWatch(
 			created: new Date(),
 			snoozedUntil: null,
 			notes,
+			...(isPublicallyTradeable !== undefined && {
```
jamesjamail: **Re-watching restores the marketplace opt-in.**

*Updated — this is much narrower now that the default is flipped to on. Keeping it for the one case that still matters.*

The conditional spread keeps `isPublicallyTradeable` when the option is omitted, and the same `update` branch sets `active: true`, so `/watch` on a previously-ended watch revives it carrying whatever was stored.

With default-on that's mostly just the product default reasserting itself, not a consent violation — and the important case works correctly: a user who explicitly passed `marketplace: false` has that `false` preserved through unwatch and re-watch, because the update branch never touches the field when the argument is absent. Same for pre-migration watches, which the migration pins to `false` and which stay out unless the user explicitly opts in.

So **no change needed here** — I'd previously suggested `unwatch` should clear the field, and that's now wrong: clearing it would fight the default rather than protect anyone.

The one thing to be deliberate about is the resulting asymmetry. Re-watching an ended item preserves its stored value; creating a watch for a *new* item gets `true` from the column default. Two watches, same user, same day, different marketplace status — defensible, but it should be a decision rather than an accident, and it's another reason the 🤝 toggle needs to show its state plainly.

(Minor, same hunk: the conditional spread is functionally identical to just writing `isPublicallyTradeable`, since Prisma already treats `undefined` as "skip this field" — which is exactly how `notes` above gets its stickiness. Not a bug, but the same behaviour is now expressed two different ways in one object.)

src/lib/commands/slashCommands/watch.ts
```diff
@@ -50,8 +53,18 @@ const command: SlashCommand = {
 				watchType: args.type.value as WatchType,
 				priceRequirement: args?.price?.value as number,
 				notes: args?.notes?.value as string,
+				isPublicallyTradeable: args?.marketplace?.value as boolean,
 			});
 
+			try {
+				await checkForMarketplaceMatches(data);
```
jamesjamail: **Marketplace work is awaited before `interaction.reply()`.**

The command never defers, so everything `checkForMarketplaceMatches` does — the match inserts plus a `client.users.send` per side of every new match, each an HTTP round trip that may also have to open a DM channel — has to finish inside Discord's ~3s ack window.

This isn't an edge case on popular items, it's the default: `isPriceCompatible` returns `true` whenever either side omits a price, so an opted-in watch with no price pairs with *every* counterpart on that item/server. When it blows the window the user sees "The application did not respond" and never gets the confirmation embed or its buttons, even though the watch saved fine.

`deferReply()` before the marketplace work fixes it in one line without making the match check fire-and-forget.

src/lib/content/messages/messageBuilder.ts
```diff
@@ -302,6 +303,61 @@ export async function watchNotificationBuilder(
...
+	if (wikiUrl) {
+		authorProperties.url = wikiUrl;
+	}
+
+	return new EmbedBuilder()
+		.setColor(getServerColorFromString(mine.server))
+		.setAuthor(authorProperties)
+		.setTitle(`Marketplace Match: ${toTitleCase(mine.itemName)}`)
+		.setDescription(
+			`Your **${mine.watchType}** watch for **${toTitleCase(mine.itemName)}** matches another trader's opposite listing. Reach out to make a deal!`,
```
jamesjamail: **The match DM labels the buyer as the seller, and vice versa.**

`watchNotification.ts:74-83` establishes the convention: a WTS-type watch treats `priceRequirement` as "a budget cap" (owner is the **buyer**), a WTB-type watch as "a minimum offer" (owner is the **seller**). The matching logic honours it correctly — `isPriceCompatible(wtb, wts)` returning `wtb.priceRequirement <= wts.priceRequirement` is seller's floor against buyer's ceiling.

The embed doesn't. Someone who created a WTS watch to buy an item is told "Your **WTS** watch ... matches another trader's opposite listing," and the actual seller is described as having a **WTB** watch. Both read backwards, inviting two people to contact each other with inverted assumptions about who's buying.

The fixture trips on it too — `messageBuilder.test.ts` builds the counterpart with `watchType: WTS` and names them `discordUsername: 'Seller'`.

Fix in copy, not the enum: derive the role and never print the raw watch type.

```ts
const theirRole = theirs.watchType === WatchType.WTB ? 'selling' : 'buying';
```

> `@Ogrelord` is selling Swiftblade of Zek for 3k — get in touch to make a deal.

In a batched digest the item is already the title, so rows shorten to `@Ogrelord — selling, 3k`.

src/lib/marketplace/marketplaceMatching.ts
```diff
@@ -0,0 +1,102 @@
...
+		const embeds = [marketplaceMatchBuilder(mine, theirs)];
+		const components = buttonRowBuilder(
+			MessageTypes.watch,
+			[false, false, false],
+			String(mine.id),
+		);
+		await client.users.send(mine.discordUserId, { embeds, components });
+	} catch (error) {
+		if (!isClosedDmError(error)) {
+			await releaseMarketplaceMatchNotificationClaim(match.id, side);
```
jamesjamail: **The claim release can swallow the original send error.**

`releaseMarketplaceMatchNotificationClaim` is awaited before `gracefullyHandleError`, and it calls `prisma.marketplaceMatch.update`, which throws P2025 if the row is gone. If a watch is unwatched concurrently, `deleteMarketplaceMatchesForWatchIds` removes the match row mid-send: the send fails, the release then rejects, and that rejection replaces the original error. `gracefullyHandleError(error, ...)` never runs, so the real send failure is never logged with its `matchId`/`side`/`watchId` context — only a bare P2025 surfaces via `Promise.allSettled`.

Wrap the release in its own try/catch, or report the original error before attempting it. This one survives into the batched-digest design unchanged, so it's worth fixing independently.

jamesjamail (commented):
## Review: marketplace matching

Five findings below, most severe first. Two are consent-related and are the ones I'd treat as blocking; the rest are ordinary correctness.

Separately, the per-pair notification model doesn't survive contact with a popular item, and that shapes several of the findings. A new opted-in watch matching 50 counterparts produces 50 sequential `MarketplaceMatch` inserts, then 100 concurrent `client.users.send` calls (50 DMs to the creator, one to each counterpart), all awaited before `interaction.reply()`. That guarantees the ack-window failure in finding 3, and rapidly opening 50 DM channels is the pattern Discord's anti-spam heuristics watch for.

### Agreed direction

Matching logic and the `MarketplaceMatch` ledger stay as written. What changes:

- **One batched digest per watch per cycle**, capped at ~15 rows with an "and N more" footer. Keep per-match claims exactly as they are and change only the send grouping — batching the claim into one `updateMany` returns a count rather than which rows were claimed, so you couldn't tell which counterparts belong in the message.
- **Matches found at creation are not DMed.** They go into the existing ephemeral `/watch` reply as a second embed, since the user is already looking at it. Drops the creation path from N+1 messages to one.
- **`/marketplace`**, no-arg and grouped by item, parallel to `/list`. `listCommandResponseBuilder`'s 25-field chunking and `packEmbedsForDiscord` already do the work.
- **Snooze covers marketplace notifications** at both watch and user level, while leaving you listed and contactable. That means dropping the snooze predicates from `findEligibleWatchesByType` and checking them at notify time instead. Worth being deliberate that a returning user then sees a backlog.
- **Governing rule:** matching stays complete; everything a user can suppress is a filter at notify/render time, never at match time. The one exception is the hard block in finding 1, which has to be mutual to mean anything.
- **A 🤝 toggle** for `isPublicallyTradeable` on the digest row (💤 ❌ ♻️ 🤝, four of five slots). No hidden mute boolean — a suppression state you can't read off the message is the same trap as finding 2.
- **Dismissal as `/marketplace hide <trader>`** with autocomplete. There are no select menus in this codebase and `interactionCreate` doesn't route them, and the button row has no room.

---

## Update: marketplace defaults to on

Reversing the opt-in posture. `isPublicallyTradeable` goes back to `@default(true)`, so a new watch is marketplace-visible unless the user opts out with `marketplace: false`.

**Existing watches are not enrolled.** The migration keeps its `UPDATE "Watch" SET "isPublicallyTradeable" = false`, so everything predating this PR stays out until its owner explicitly opts in. Default-on applies going forward only.

Code changes this implies:

- `schema.prisma` — `isPublicallyTradeable Boolean @default(true)`, and update the comment above it, which currently reads "opt-in, off by default"
- `upsertWatch` — the `create` branch's `isPublicallyTradeable ?? false` becomes `?? true` (or drop the field and let the column default apply)
- `marketplaceOptions` — the description still says "off by default"; it should say the opposite
- The migration's `UPDATE` line stays, and its comment should explain that it is now deliberately holding existing rows back from a default that would otherwise enroll them

Two consequences worth being deliberate about, rather than discovering later:

**The toggle is now the consent control.** With opt-in, a user chose to be listed and the 🤝 button was a convenience. With default-on, it is the only thing standing between a user and unsolicited DMs from strangers, so its state has to be legible on every digest — not just the icon variant, but spelled out in the footer.

**Old and new watches will disagree.** A user's pre-existing watches are pinned `false` while anything they create after the migration defaults `true`. That is the correct outcome of "new watches only", but it means the same person can hold two watches with opposite marketplace status and no obvious reason why. `/marketplace` showing which of your watches are listed would go a long way.

Finding 2 below has been updated — it is much narrower under default-on, and my earlier suggestion that `unwatch` should clear the field is withdrawn.

src/prisma/migrations/20260916205135_marketplace_matching/migration.sql
```diff
@@ -0,0 +1,29 @@
+-- Marketplace matching: pairs an active WTB watch with an active WTS watch for
+-- the same item/server when both sides are marketplace-visible
+-- (isPublicallyTradeable, which defaults to true for new watches).
+--
+-- Every watch that exists before this migration was created without the user
+-- knowing it would put their handle in strangers' DMs. Holding them back here
+-- is deliberate: without it they would all be enrolled by the default. Their
+-- owners can opt in explicitly with /watch marketplace:true.
+UPDATE "Watch" SET "isPublicallyTradeable" = false;
```
rm-you, line 9: **[P1] Preserve the existing-user opt-out during a direct PostgreSQL-to-SQLite upgrade**

This UPDATE handles an existing SQLite database, but misses another supported startup path. `docker-entrypoint.sh` runs all migrations first and then `import-postgres.js`. On a fresh destination this UPDATE affects zero rows. The importer subsequently copies `Watch.isPublicallyTradeable` unchanged from PostgreSQL, where the legacy default was already true even though marketplace matching did not exist. Those old watches can therefore start sharing their owners' handles with matching traders, contrary to the agreed "existing watches are not enrolled" behavior.

I reproduced the migration/import ordering with the actual SQLite migrations and a synthetic legacy row: after migration followed by insertion of the legacy true value, the resulting watch is listed. I also traced the importer's column-copy path; this reproduction did not connect to a live PostgreSQL server.

**Requested fix:** explicitly normalize legacy PostgreSQL Watch rows to unlisted during this one-time import, or use an equivalent version-aware migration after import. Keep new-watch defaults enabled. Do not blanket-reset existing SQLite consent on every restart.

**Regression coverage:** upgrade a legacy PostgreSQL fixture with active watches whose old value is true into an empty SQLite destination; after startup/import they must be false, while a newly created watch defaults true. Also preserve the existing SQLite-upgrade behavior and verify restart/import idempotence.

src/prisma/dbExecutors/marketplace.ts
```diff
@@ -0,0 +1,501 @@
...
+		},
+		include: matchWithWatchesInclude,
+		orderBy: { id: 'desc' },
+	});
+
+	for (const match of matches) {
+		for (const side of MATCH_SIDES) {
+			const { mine, theirs } = sidesOfMatch(match, side);
+			const view = viewByWatchId.get(mine.id);
+			if (!view || !isListed(theirs)) continue;
```
rm-you, line 367: **[P2] Revalidate current prices before exposing an existing match**

`isPriceCompatible()` is applied when creating candidate pairs, but persisted pairs pass this rendering path based only on active/listed status and suppression filters. Updating a watch's price does not remove or invalidate its match rows.

For example, a seller watching WTB initially asks 500 and matches a buyer watching WTS with a 1,000 ceiling. The seller later changes the floor to 2,000. The saved pair is still returned as a current match, including to the notification path if that side has not yet been notified. A focused reproduction with floor 2,000 and ceiling 1,000 returned one counterpart despite `isPriceCompatible()` returning false.

**Requested fix:** recheck `isPriceCompatible(match.wtbWatch, match.wtsWatch)` using the freshly loaded watches before adding counterparts to either view. Retaining the ledger and filtering on read fits the current architecture; a price edit should not require deleting notification history.

**Regression coverage:** create a compatible pair, change either side's price to make it incompatible, and verify it disappears from both listing and pending-digest views. Restore compatibility and verify it becomes visible again without spuriously resetting already-delivered notification history. Keep unset-price behavior negotiable.

src/lib/content/messages/messageBuilder.ts
```diff
@@ -398,6 +504,125 @@ export function listCommandResponseBuilder(
...
+	const name = `\`${marker} ${toTitleCase(watch.itemName)}\` | \`${tradeRoleOf(watch).role}\``;
+
+	let value: string;
+	if (!watch.isPublicallyTradeable) {
+		value =
+			'Not listed: other traders cannot see this watch. Run /watch for this item again with marketplace set to true to list it.';
+	} else if (counterparts.length === 0) {
+		value = 'No matching traders yet.';
+	} else {
+		const shown = counterparts.slice(0, MARKETPLACE_FIELD_ROW_LIMIT);
```
rm-you: **[P2] Make the results omitted from a digest accessible through /marketplace**

The digest displays 15 counterparts and says "Use /marketplace to see them all," but this command then slices each watch to eight counterparts. There is no continuation field, page, or other listing option for the omitted traders. The existing packing code only distributes the already-truncated fields; it cannot recover those rows.

With 20 synthetic counterparts I got 15 in the digest and only eight in the advertised fallback. Further, the sender claims all counterpart notifications, including those omitted from the digest, so those omitted rows cannot be expected to appear in a later notification.

**Requested fix:** keep the digest compact, but let /marketplace enumerate every visible counterpart through continuation fields/embeds/messages or actual pagination. Respect the existing per-field, per-embed, and per-message limits. Merely changing "see them all" to different wording would leave users unable to access matches.

**Regression coverage:** use more than 15 counterparts for one watch and multiple watches/servers. Assert every eligible trader is reachable, no trader is silently dropped, and every resulting Discord payload remains within limits.

src/lib/commands/slashCommands/watch.ts
```diff
@@ -44,26 +51,39 @@ const command: SlashCommand = {
...
+				marketplaceEmbed = await checkForMarketplaceMatches(data);
+			} catch (error) {
+				await gracefullyHandleError(error, interaction, command, {
+					watchId: data.id,
+					phase: 'marketplaceMatching',
+				});
+			}
+
 			const embeds = [watchCommandResponseBuilder(data)];
+			if (marketplaceEmbed) embeds.push(marketplaceEmbed);
```
rm-you: **[P2] Show saved marketplace consent and provide the advertised toggle in the creation response**

The optional second embed comes from `marketplaceDigestBuilder()`, whose footer tells the user "Click 🤝 to unlist this watch." This reply still uses `MessageTypes.watch`, which has only 💤, ❌, and ♻️. The advertised consent control is absent.

There is a related empty-state problem: `checkForMarketplaceMatches()` returns undefined when no counterpart exists, and the ordinary watch confirmation does not display marketplace visibility at all. A new watch is therefore listed by default without that confirmation explaining it. This matters because pre-existing watches can be unlisted while new watches are listed.

**Requested fix:** always show the persisted listing state on the ordinary watch confirmation, even when there are no matches, and provide an actual list/unlist control or an equally clear actionable alternative. Where the footer promises a 🤝 button, supply that button. Ensure the handlers re-render the appropriate watch/preview state rather than making the confirmation or status stale after another button is clicked. Reuse existing components where possible; no new interaction framework is needed.

**Regression coverage:** new default-on watch with zero matches; explicit marketplace:false; an existing unlisted watch updated with the option omitted; and a creation preview with matches. Verify the copy matches persisted state and every advertised button exists and operates correctly.

src/lib/marketplace/marketplaceMatching.ts
```diff
@@ -0,0 +1,146 @@
...
+export async function checkForMarketplaceMatches(
+	watch: Watch,
+): Promise<EmbedBuilder | undefined> {
+	await matchNewWatchToMarketplace(watch);
+
+	const view = await getMarketplaceViewForWatch(watch.id);
+	if (!view || view.counterparts.length === 0) return undefined;
+
+	// 	they are about to be shown these, so the digest must not repeat them
+	await claimCounterparts(
```
rm-you: **[P2] Recover preview notification claims if building or delivering the creation reply fails**

This function claims all unnotified counterparts before constructing the embed and before its caller runs `interaction.editReply()`. Neither a rendering exception nor a failed edit releases those claims. In contrast, the DM path explicitly releases claims for retryable delivery failures.

I injected a failed creation reply with the real command/matching flow and mocked database/Discord boundaries. The recipient-side timestamp remained set, and the following sweep sent no notification. As a control, clearing that timestamp made the same sweep deliver the match. The pair remains discoverable manually, but automatic delivery of those matches is suppressed until something resets the ledger.

**Requested fix:** make preview claims part of an explicit delivery lifecycle. Construct/validate the payload before acquiring claims where possible, return the exact slots this invocation successfully claimed, and keep them committed only after a successful reply; release this invocation's claims on definite rendering/delivery failure. Keep concurrency protection against the periodic sweep, and avoid clearing another invocation's claims. Also unwind already-acquired slots if claiming a later counterpart throws. Moving a blanket "mark notified" after delivery without concurrency protection would introduce a different duplicate-delivery race.

**Regression coverage:** failed embed construction, failed editReply, partial claim failure, successful reply, and a concurrent sweep. Failed delivery must remain retryable; successful delivery must not be repeated by the next sweep.

src/lib/content/messages/messageBuilder.ts
```diff
@@ -302,6 +307,107 @@ export async function watchNotificationBuilder(
...
+	}
+
+	if (wikiUrl) {
+		authorProperties.url = wikiUrl;
+	}
+
+	return new EmbedBuilder()
+		.setColor(getServerColorFromString(watch.server))
+		.setAuthor(authorProperties)
+		.setTitle(
```
rm-you: **[P2] Bound the final marketplace title, including its prefix and server suffix**

The schema allows a 255-character item name, and custom item names are accepted by /watch. The title adds "Marketplace: " and the server suffix without truncation. A synthetic 255-character name makes the installed Discord embed builder throw rather than render.

This affects both creation previews and background digests. In the creation path it also combines with the early-claim issue above; in the DM path it can cause recurring failed delivery attempts.

**Requested fix:** reserve room for the prefix/suffix and truncate the displayed item name, or truncate the final title safely to Discord's 256-character limit. Keep the actual stored item name unchanged.

**Regression coverage:** longest accepted custom names, including Unicode, across the server variants. Validate the complete serialized payload, not just its description/fields. A 255-character item name must not prevent a valid watch confirmation or digest.

src/prisma/dbExecutors/watch.ts
```diff
@@ -205,6 +217,8 @@ export async function unwatch(metadata: MetadataType) {
 			snoozedUntil: null, //	unwatching should remove any snooze
 		},
 	});
+	await deleteMarketplaceMatchesForWatchIds([watch.id]);
```
rm-you, line 220: **[P2] Keep notification abuse protection independent of watch/match deletion**

Deleting every pair here also deletes the evidence that the other trader was already notified. The same user can end and restore/re-create a watch for a popular item, after which the matcher creates fresh unnotified pairs and notifies the same counterpart again. The same issue applies to the other unwatch entry points.

I reproduced this with a temporary real SQLite database, all PR migrations, actual upsert/unwatch/matching/sweep code, and mocked Discord delivery:

1. Two synthetic users create complementary watches; the recipient receives one digest.
2. A second unchanged sweep sends nothing.
3. The counterparty ends and restores their watch.
4. The next sweep sends the same recipient another digest for the same trader/item.
5. An explicit recipient-side trader block prevents subsequent delivery, confirming that blocking works but requires the recipient to intervene.

This is bounded by the normal five-minute sweep and one digest per recipient watch per cycle; creation does not immediately DM other users. Nevertheless, one sender can repeatedly re-trigger notifications across common items and multiple recipients. The 3s /watch and 10s /unwatch command cooldowns do not address a five-minute repeat cycle, and button interactions do not use that slash-command cooldown path. Re-running /watch without ending it preserves the pair and is not the same reset path.

**Requested fix:** retain a persistent notification throttle/deduplication record keyed by stable recipient/counterparty identity plus the relevant item/server/direction, independent of ephemeral match IDs and watch deletion. A counterparty ending, restoring, or replacing a watch must not reset the recipient's quiet period. Define a reasonable repeat-notification policy so genuine relisting can eventually notify again. Add a bounded per-recipient delivery budget/backoff across watches so creating many different matching watches cannot produce an unbounded burst. This can preserve the agreed per-watch digest format; it does not require redesigning the UI into per-user digests. Recipients should still be able to discover currently valid matches through /marketplace even when notifications are throttled.

**Regression coverage:** unchanged sweeps, repeated end/restore, genuine replacement watch IDs, many matching items targeting one recipient, cooldown expiry, failed-send retry, and block/hide behavior. Include an unchanged recipient as the control: the sender's lifecycle operations alone must not keep rearming notifications.

rm-you (changes requested):
## Review of 7aecebb

**Disclosure: this review was primarily AI-driven, using Codex for source inspection and isolated local reproductions, and is being posted at the repository collaborator's explicit request. It is not a claim of independent human verification or a live Discord/production test.**

The overall implementation fits the existing TQB structure: typed Prisma models, a database-backed pair ledger and uniqueness constraint, conditional notification claims, separate view/building code, and substantial unit coverage. The current revision addresses the earlier review's major direction changes: ephemeral creation previews, per-watch digest batching, mutual trader blocks, and snooze filtering at delivery time. I am not requesting a framework rewrite or a change from the agreed per-watch digest model.

I am requesting changes for the seven specific findings attached inline:

1. **P1:** a direct legacy PostgreSQL upgrade can enroll existing watches despite the migration's intended opt-out.
2. **P2:** persisted matches remain visible/notifiable after incompatible price changes.
3. **P2:** /marketplace cannot reveal the counterparts omitted from the digest.
4. **P2:** creation responses omit the advertised consent toggle, and the no-match confirmation hides listing status.
5. **P2:** failed preview construction/delivery leaves notification claims permanently consumed for those rows.
6. **P2:** valid maximum-length custom item names overflow the marketplace title.
7. **P2:** a counterparty can reset notification history through end/restore and repeatedly notify the same recipients.

Each inline comment describes the trigger, requested behavior, and regression coverage. These are distinct from the follow-ups below.

## Reliability and UX follow-ups

These are recommendations, not additional reproduced production incidents or separate blockers for this review:

- **Atomic lifecycle cleanup:** unwatch currently updates the watch and deletes its matches in separate operations. If deletion fails, the watch is inactive but old notification history remains, so restoration may not behave as expected. Make the lifecycle update/cleanup atomic. A transaction alone does not cover a sweep that loaded an eligible candidate before deactivation and inserts it afterward: recheck eligibility at the write boundary, or choose an equivalent lifecycle/version strategy. Current read-time active/listed filtering is useful and means a leftover row does not automatically imply continued exposure. Coordinate this with finding 7: retain abuse-throttling history even if live match rows are cleaned up.
- **Acknowledge marketplace buttons early:** the routing path looks up metadata, then mutates the database and rebuilds the view, before calling interaction.update(). Slow database work can exceed Discord's [three-second acknowledgement deadline](https://github.com/discord/discord-api-docs/blob/main/developers/interactions/receiving-and-responding.mdx). For marketplace actions, deferUpdate before the first potentially slow lookup and then editReply; update missing-entity/error branches consistently. A deferred/fake-clock test can prove acknowledgement precedes delayed work. This is a source-confirmed timeout risk, not a measured live timeout.
- **Narrow the unlisting promise:** yourWatchIsNoLongerListed says other traders cannot "contact you about it." Unlisting prevents future marketplace discovery; it cannot retract identities already sent or prevent direct Discord messages. Suggested wording: "This watch is no longer listed and will not appear in future marketplace results or notifications. Traders who already saw your Discord identity may still contact you." The existing block confirmation is already more accurately scoped to matching/sharing handles.
- **Bound background work:** candidate arrays are a WTB x WTS product within each item/server bucket, and pending-watch delivery uses an unbounded Promise.allSettled. The non-overlapping interval guard, grouping, uniqueness checks, and Discord REST queue are useful safeguards; they do not bound application memory or database fan-out. Use bounded delivery concurrency, and consider incremental candidate processing based on measured watch counts. I have not established a production scale at which the current code fails. Concurrency limits alone also do not replace the recipient notification budget requested in finding 7.

## Deliberately not treating these as bugs

- Multiple notifications to a user with several watches follow the explicitly agreed **one digest per watch per cycle** design. The abuse concern is the missing persistent repeat/burst control, not the digest grouping itself.
- The claim that /unblocktrader necessarily cannot address someone who left the guild is not established. It accepts a Discord user, rather than requiring a guild member; user options support specifying outside-guild users by ID. Stored-block autocomplete would improve discoverability, but is not required to resolve a demonstrated inability to unblock.
- Default-on for newly created watches is the agreed product choice. The requested corrections concern disclosure, functional controls, and preserving the exception for legacy users.

## Validation and limits

- Verified current head is **7aecebbf8be4f662a36d39871d376a4de908fa9b**.
- Generated the Prisma client; typechecking passed.
- Re-ran the existing unit suite under Node 24: **94 files / 849 tests passed**.
- Six focused local regression checks reproduced findings 1–6. Those use mocked boundaries where appropriate; the PostgreSQL finding uses the real SQLite migration sequence plus a synthetic legacy import row and source tracing, not a live PostgreSQL import.
- Separately reproduced finding 7 using a real temporary SQLite database and actual lifecycle/matching code, with Discord sends mocked. Unchanged sweeps and explicit blocking were checked as controls.
- No production data, real credentials, real Discord messages, or game-server traffic were used. The Docker-based integration suite and live Discord UI were not run. GitHub currently reports no checks on this head.
- Review-only fixtures were kept local; no commits or code changes are included in this review.

rm-you:
Reviewing my own AI review:

"Acknowledge marketplace buttons early": I've been bit by this so many times, I really recommend always doing a defer+edit in almost every case when possible.

"The claim that /unblocktrader...": actually I think the AI is wrong, I've run into this problem where someone leaves and I can't auto fill them as a discord user, but also it doesn't really MATTER if you can unblock them if they aren't in the server anymore? Idk, granted it is just more sane to do a custom auto fill with only actual blocked users listed, which doesn't require setting the field's type to discord user. Also, is that a new command? Need to look at our command list again, I feel like we're maybe cluttering our tab-complete for "unblock" but I don't remember offhand

